PR #2, opened 2012-07-23
GATEWAY-410: remove links to NSFW web site in ivy.xml files

ccabot:
At some point, the maven search site indicated in the ivy.xml files
changed to no longer be a search engine for Maven artifacts.  It's now
NSFW.  This patch removes the comments that link to it.


bhumphrey:
Looks like http://www.mavensearch.net/ is no longer a valid site. This commit removes this site from the commits in the ivy.xml.

PR #6, opened 2012-09-14
Transaction Logging - GW2159, GW2565

jsmith2:
Changes for Gateway-2564 and Gateway-2159.  Created DAO for transaction handler and added logging option and also changed some of the logic of the transaction handler for adding reply to and not adding a transaction log to every message.


zmelnick:
Unit tests are occurring in GW2739. Will review and merge after those are written.

PR #7, opened 2012-09-17
Gateway 2734

akong:
Modified build script to support the copying of wsdl and schemas into the project wars.  This is only done for PD and it is optional (for now).  Also deleted all local wsdl and schemas in the PD wars.


dharley-5am:
looks good to me.

PR #8, opened 2012-09-17
Gateway 1284

zmelnick:
Unit tests for Orchestration Implementation of AD, plus a bunch of cleanup code.


msweaver:
Looks good to me.

PR #12, opened 2012-09-19
GATEWAY-2672: Message from CONNECT to DIRECT using James mail server

arathee:


msweaver:
Let's not merge this code. I have created another pull request to put this code into a branch. Please use that CR/PR instead. 

https://github.com/CONNECT-Solution/CONNECT/pull/16

PR #23, opened 2012-09-25
fixes issue: GATEWAY-2769 (hibernate should use container-managed dataso...

dharley-5am:
giving access to brian, since he is blocked by this issue on websphere. this is not ready to be pushed to CONNECT_integration, so no need to review.


Product/Production/Services/PatientDiscovery/Core/PatientDiscoveryCore/testMPI.xml
```diff
@@ -1,4 +1,4 @@
```
bhumphrey, line 0: i don't think you want to check this in..

dharley-5am, line 0: no, i didn't, thanks for catching.  i will revert.

PR #26, opened 2012-09-25
Fixed problems with RelatesTo for Transaction Logging

jsmith2:
Was correlating on replyto which is incorrect, now correlating on relatesto.  Also, added ability for multiple correlations on relatesTo since it is a repeatable element, allowed for this both in the database and the logger.

### DIFF
--- a/Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/logging/transaction/TransactionHandler.java
+++ b/Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/logging/transaction/TransactionHandler.java
@@ -63,7 +63,7 @@ public class TransactionHandler implements SOAPHandler<SOAPMessageContext> {
     private static Log log = LogFactory.getLog(TransactionHandler.class);
     private static final String WSA_NS = "http://www.w3.org/2005/08/addressing";
     private static final String MESSAGE_ID = "MessageID";
-    private static final String REPLYTO_ID = "ReplyTo";
+    private static final String RELATESTO_ID = "RelatesTo";
      
     protected Log getLogger(){
     	return log;
@@ -79,7 +79,7 @@ public class TransactionHandler implements SOAPHandler<SOAPMessageContext> {
     	
     	String messageId = null;
         String transactionId = null;
-        String replyToId = null;
+        String relatesToId = null;
 
         SOAPMessage soapMessage = context.getMessage();
         SOAPPart soapPart = soapMessage.getSOAPPart();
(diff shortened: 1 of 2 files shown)


Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/logging/transaction/TransactionHandler.java
```diff
@@ -104,20 +104,36 @@ public boolean handleMessage(SOAPMessageContext context) {
...
+                                    		relatesBuffer.append(", ");
+                                    	}
+                                    	relatesBuffer.append(transactionId);
+                                    	count++;
+                                    }//else transactionId is already persisted
+                        		}
+                        	}
+                		}
+                		if(relatesBuffer.length() > 0){
+                			transactionId = relatesBuffer.toString();
```
bhumphrey, line 133: why are you doing this?

jsmith2, line 133: After talking to Matt W., it was discovered that the relatesTo element is the correct element to correlate off of.  Matt T. also pointed out that the relatesTo element is a repeatable element.  So I added the logic to check the for multiple relates to elements and see if they have a transactionID.  I also allowed for multiple transaction IDs to be logged for a message that has the multiple relatesTo messages with multiple transactionID correlations.

bhumphrey, line 133: so you could end up with aaa,bbb,ccc ddd in the log files? meaning this message was related to three transactions? While that is correct I think it clutters the log files. Adding a single transaction id would be sufficient. The database will hold the many-to-one relationship.

jsmith2, line 133: Ok.  I think this might be pretty edge case.  Not sure if it will clutter it too much.

Would be:
[ log output .....
(transaction ID) (transactionID) .....,
messageID

Might make one line a little longer, not sure if I would consider that clutter and would add useful info.  Easy change to revert though so i can go either way.

msweaver, line 133: I could go either way on this one too. I don't think there are going to be very many instances of there being multiple transaction ids resulting from multiple relatestos. Just logging one transaction is more simple, but seeing as you had already done the work, I don't think there is enough reason to revert.

This is one of those things where either way you had done it, I don't think there is enough reason to advocate spending more time on it. Especially on such a fringe case.

PR #28, opened 2012-09-26
GATEWAY-2714: mvn based build

ghost:


5AMSolutions:
Being that I'm new to CONNECT & github, what can I do to merge this into the destination?

PR #32, opened 2012-09-26
GATEWAY-2306: fixed issue with parsing saml assertion from dom, also rem...

mlandis5am:
...oved deprecated perf mgr calls


mlandis5am:
That was from this previous commit: c855a3f15dc2367b7690000135b6d65593015543.  This class would probably benefit from an exception handling strategy.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/openSAML/extraction/OpenSAMLAssertionExtractorImpl.java
```diff
@@ -480,32 +490,26 @@ private void populateSubjectRole(final Attribute attribute, final AssertionType
 
         for (Map.Entry<QName, String> entry : attributeMap.entrySet()) {
             QName key = entry.getKey();
-            if (key.getNamespaceURI().equals(NhincConstants.HL7_NS)) {
```
bhumphrey, line 483: I think you still want to check the namespace.

PR #33, opened 2012-09-27
GATEWAY-2810

msweaver:
My work so far on GATEWAY-2810. So far this branch contains two commits:
First commit fixes the CONNECT internal "spec version registry", Updates the callbackhandler "decider" code and puts a lot of unit tests around the purposeof/purposefor decider.


akong:
Just a general question.  Are we going to have a new PatientDiscovery_20 war?  And will the OutboundPatientDiscoveryFactory get updated to handle G1 as well?

msweaver:
So I intend to keep only the PatientDiscovery_20.war, these changes don’t affect the receiving side.

Since the messages are still handled the same way up until the callback handler I had a different approach than adding a G1 strategy etc to the Factory. I was thinking I could just have one “common” impl that got called for g0 or g1. I’ll put that into it’s own commit so you can see it when it’s done.

msweaver:
@akong commit 2597b31 should address your comment.

msweaver:
NhinEndpointManager returns LEVEL_g1 as a default even for specversion 1.0 Patient Discovery Deferred - which is not available in 1.0 of Patient Discovery. Any thoughts on how to better handle this case?

akong:
Is that by design?  If orchestration doesn't understand the version number (ex. PD 3.0), it'll try its best to route it to the highest version supported (g1).  I think it'll be cleaner if NhinEndpointManager returns an UNKNOWN_GATEWAY_LEVEL enum and have the ConnectionManager handle the return of the highest version supported url.

PR #41, opened 2012-10-01
Maven Build

ghost:


mlandis5am:
I sanity checked a few of the files from previous pull requests and the changes are reflected in the pull request.

akong:
Sanity checked that large payload feature was included in the merge.

dharley-5am:
all the changes i have pushed recently are ant-related and would be superceded by these changes, except the roll-back of the hibernate datasource usage, which steve has already merged-in, so things are good-to-go from my perspective...will need to work through problems when it lands.

PR #54, opened 2012-10-04
setting mutli-module for application specific building

bhumphrey:


Product/Production/CONNECT/ear/pom.xml
```diff
@@ -0,0 +1,862 @@
```
5AMSolutions, line 0: Could you do a git mv to preserve file history???

PR #55, opened 2012-10-04
Clarification in readme

matthewtiller:


dharley-5am:
only commit 88799d0 should go though, right?  that commit looks good, thanks for the clarification.

PR #74, opened 2012-10-10
fixes issue: GATEWAY-2860 (be default activate PD,DQ,DR,DS,AD so that th...

dharley-5am:
fixes issue: GATEWAY-2860 (be default activate PD,DQ,DR,DS,AD so that they get built with a naked mvn clean install)


dharley-5am:
GUI is not active by default:

```
    <profile>
        <id>GUI</id>
        <activation>
            <activeByDefault>false</activeByDefault>
        </activation>
```

PR #89, opened 2012-10-15
Gateway 2769 - switch from c3po to container managed pool

mlandis5am:
Notes:
- "mysql-connector-java-5.1.10.jar" must be copied into the following folder:
  
  ```
  ${GLASSFISH_HOME}/domains/domain1/lib/ext
  ```
- we'll need to drop that jar in place for other containers when they are added, like websphere.
- connection pool sizes would likely need to be tweaked based on a particular deployment.


dharley-5am:
to integrate with the embedded glassfish, these changes will also need to be made:
1) add datasource/connection pools/resource-refs to Product/SoapUI_Test/ValidationSuite/src/test/resources/domain1/config/domain.xml
2) modify the Product/SoapUI_Test/pom.xml to copy the mysql driver jar from maven repo to the embedd glassfish's domain lib dir (${project.build.directory}/domain1/lib prior to glassfish startup)

PR #96, opened 2012-10-18
GATEWAY-2786: Implemented large payload feature to RD

akong:
A few things to look for:
1. The legacy implementation of DocRetrieve (especially the DocRetrieveAggregator) has been removed from the code base.
1. MTOM is now explicitly enabled and per CXF design, it will always return if it is labelled as optional (which our WSDLs do).  This means that the SoapUI test for DR has to have its schema compliance turned off.  The reason is that the xop:include for MTOM is inserted inside the <Document> element and is not strictly xsd compliant.  This is a known issue of most schema validators including the ones SoapUI uses.
2. DocRepository NoOp adapter has been modified such that we can let it return a response with fields populated from the Spring Proxy config.


Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/nhinclib/NullChecker.java
```diff
@@ -54,7 +55,19 @@ public static boolean isNullish(List value) {
         return result;
     }
 
-    public static boolean isNotNullish(List value) {
+    public static boolean isNotNullish(List<?> value) {
```
bhumphrey, line 50: this should just be 

```
public static boolean isNotNullish(collection value) {
   return value != null && !value.isEmpty();
}
```

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/largefile/TempFileInputStream.java
```diff
@@ -0,0 +1,31 @@
...
+    
+    public TempFileInputStream(File file) throws FileNotFoundException {
+        super(file);
+        this.file = file;
+    }
+
+    @Override
+    public void close() throws IOException {
+        super.close();
+        file.delete();
```
bhumphrey: i think this needs to be:

```
try { 
  super.close();
} finally {
  file.delete();
}
```

bhumphrey:
Having the noop version do something isn't that kind of defeating the no operation idea?

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/nhinclib/NullChecker.java
akong, line 50: Yeah, that's better.  I'll fix it.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/largefile/TempFileInputStream.java
akong: Good catch.

akong:
I was debating whether I should create a new class or whether to add it to the NoOp implementation.  The change that I made was fairly minimal in processing, so I convinced myself to put it to NoOp. 

But it sounds like it's better to just move it to a different class and maybe add a few more fields to it.  I'll go ahead and do that.

PR #100, opened 2012-10-18
Gateway 2927

alameluchidambaram:
Remove AD Delay of 15 sec


Product/Production/Services/AdminDistributionCore/src/main/java/gov/hhs/fha/nhinc/admindistribution/nhin/NhinAdminDistributionOrchImpl.java
```diff
@@ -68,7 +68,7 @@ public void sendAlertMessage(EDXLDistribution body, AssertionType assertion) {
         // we were hanging on the next webservice call.
         // sleep allows Glassfish to catch up. Only applies to one box (dev)
         // setups. Please refer to the CONNECT 3.1 Release Notes for more information.
-        this.checkSleep();
+       // this.checkSleep();
```
mlandis5am: Not sure if this was intended to be checked-in, if so I think we should just remove it instead of commenting it out.

mlandis5am:
finished reviewing.

PR #106, opened 2012-10-21
GATEWAY-2898: working websphere

msweaver:
CONNECT can be installed on WASCE using the instructions found here: https://developer.connectopensource.org/display/CONNECT40/Websphere+EAR+Deployment


Product/Production/Adapters/DocumentQuery_a0/src/main/webapp/WEB-INF/web.xml
```diff
@@ -12,7 +12,7 @@
         <url-pattern>/*</url-pattern>
     </servlet-mapping>
   
-    <security-constraint>
+    <!--security-constraint>
```
bhumphrey, line 15: this should be removed.

Product/Production/Gateway/DocumentQuery_20/src/main/webapp/WEB-INF/web.xml
```diff
@@ -17,7 +17,7 @@
         <url-pattern>/*</url-pattern>
     </servlet-mapping>
 
-    <security-constraint>
+    <!-- security-constraint>
```
bhumphrey, line 20: this security constraint needs to be removed

PR #109, opened 2012-10-22
explaining how user can specify mysql root user\'s password in non-default cases

dharley-5am:


README.md
```diff
@@ -180,32 +197,32 @@ Testing
 ###Run the Validation Suite as part of install
 At the end of the mvn install process, an embedded GlassFish instance will start and the Validation Suite will run against it:
         $ cd <CONNECT_CLONE_DIR>/Product/SoapUI_Test/ValidationSuite
-        $ mvn clean install
+        $ mvn clean install -Dmysql.root.password=<password>
+
+###Run the Validation Suite via Maven script
+The Validation Suite can be run via a Maven script against a standalone installation of the application server:
+        $ cd <CONNECT_CLONE_DIR>/Product/SoapUI_Test/ValidationSuite
+        $ mvn verify -Dstandalone -Dproperties.dir=<applicaiton server configuration dir> -Dmysql.root.password=<password>
```
tparnell: Why not just put all the properties into the list below (starting at line 207?)

README.md
```diff
@@ -58,15 +65,20 @@ OSX / Linux Users:
 Everyone:
 
         $ cd <CONNECT_CLONE_DIR>
+
+        If using default CONNECT MySQL root user password "NHIE-Gateway":
```
tparnell: This line repeats info from 48-54.  I don't think it's worth repeating every time there is a maven command line to tell the user about this option.

README.md
```diff
@@ -58,15 +65,20 @@ OSX / Linux Users:
...
         $ mvn clean install
 
+        If using non-default CONNECT MySQL root user password:
```
tparnell: Remove.

README.md
```diff
@@ -79,11 +91,16 @@ and choose the clone repo directory (\<CONNECT_CLONE_DIR\>), e.g. CONNECT. You m
 * \<CONNECT_CLONE_DIR\>/Product/Production
 
 ####Building an ear
-All services profiles are active by default, so to build an ear containging all services, just execute:
+All services profiles are active by default, so to build an ear containing all services, using the default MySQL root password, just execute:
 
         $ cd <CONNECT_CLONE_DIR>
         $ mvn clean install
 
+To build an ear containing all services, with non-default MySQL root user password, just execute:
```
tparnell: Similar to above, I don't think all this repetition is necessary.

README.md
dharley-5am: good idea, thanks.

PR #113, opened 2012-10-23
ignoring snapshots from forgerock

bhumphrey:
this should prevent errors like 

```
Caused by: org.apache.maven.artifact.resolver.ArtifactNotFoundException: Failure to find org.connectopensource:DBScripts:jar:
4.0.0-SNAPSHOT in http://repo.forgerock.org was cached in the local repository, resolution will not be reattempted until the
update interval of forgerock has elapsed or updates are forced
```

### DIFF
--- a/Product/pom.xml
+++ b/Product/pom.xml
@@ -370,6 +370,9 @@
             <id>forgerock</id>
             <url>http://repo.forgerock.org</url>
         </repository>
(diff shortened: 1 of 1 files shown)


dharley-5am:
looks good to me...although i have never even noticed the problem this is fixing.  merging in.

PR #117, opened 2012-10-24
GATEWAY-3020 update DS and DS Deferred ValidationSuite assertions

matthewtiller:


Product/SoapUI_Test/ValidationSuite/EntityValidation-soapui-project.xml
```diff
@@ -5,7 +5,7 @@
 		<con:settings/>
 
 		<con:endpoints>
-		<con:endpoint>http://localhost:${HttpDefaultPort}/Gateway/DocumentQuery/2_0/EntityService/EntityDocQueryUnsecured</con:endpoint></con:endpoints>
+		<con:endpoint>http://localhost:8080/Gateway/DocumentQuery/2_0/EntityService/EntityDocQueryUnsecured</con:endpoint></con:endpoints>
```
bhumphrey, line 8: does this mean we can't override them anymore? that it will always be localhost and 8080

matthewtiller, line 8: No, it does not mean that. The override still works. This change means that the tests will work in the soapUI gui, which is important for developing the tests.

PR #118, opened 2012-10-24
GATEWAY-2906: Working weblogic

msweaver:
All gateway services except HIEM are working on Weblogic.


Product/Production/Deploy/pom.xml
```diff
@@ -39,5 +39,14 @@
                 <module>websphere</module>
             </modules>
         </profile>
+        <profile>
+            <id>weblogic</id>
+            <modules>
+                <module>ear</module>
+                <module>glassfish</module>
```
bhumphrey, line 46: we don't need glassfish for the weblogic profile.

msweaver, line 46: If you don't include gfish builds from the top level fail embedded testing.

Brian Humphrey notifications@github.com wrote:

In Product/Production/Deploy/pom.xml:

> @@ -39,5 +39,14 @@
>                  <module>websphere</module>
>              </modules>
>          </profile>
> -        <profile>
> -            <id>weblogic</id>
> -            <modules>
> -                <module>ear</module>
> -                <module>glassfish</module>

we don't need glassfish for the weblogic profile.

PR #120, opened 2012-10-26
GATEWAY-2876: Added fix for CXF timestamp validation.  Exposed CXF times...

akong:
...tamp configuration.


Product/Production/Gateway/AdminDistribution_10/src/main/resources/CONNECT-interceptor-beans.xml
```diff
@@ -5,5 +5,6 @@
```
bhumphrey: CONNECT-interceptor-bean.xml can be centralized to CONNECTCoreLib

Product/SoapUI_Test/ValidationSuite/src/test/resources/domain1/config/domain.xml
```diff
@@ -317,6 +317,7 @@
         <jvm-options>-Dcom.sun.enterprise.security.httpsOutboundKeyAlias=gateway</jvm-options>
         <jvm-options>-DSERVER_KEY_ALIAS=gateway</jvm-options>
         <jvm-options>-DCLIENT_KEY_ALIAS=gateway</jvm-options>
+        <jvm-options>-Dcom.sun.grizzly.writeTimeout=30000</jvm-options>
```
bhumphrey, line 320: what about websphere and weblogic?

Product/Production/Gateway/AdminDistribution_10/src/main/resources/CONNECT-interceptor-beans.xml
akong: Ok.  I'll see what I need to do to consolidate all of it to CoreLib.

Product/SoapUI_Test/ValidationSuite/src/test/resources/domain1/config/domain.xml
akong, line 320: The configuration above is glassfish only.  I felt that it was beyond the scope of this effort (i.e. trying to find all the configuration that may affect large payloads for all servers).  I am updating the wiki to provide a guide to configure glassfish which they can use as an aid for configuring websphere/weblogic.  With that said, I believe we are doing performance testing with webpshere/weblogic and large payloads.  We can update then the instructions/config files if we see anything that will alleviate whatever bottlenecks that may pop up.

PR #121, opened 2012-10-29
Connect integration

zmelnick:


Product/Production/Services/PatientDiscoveryCore/testMPI.xml
```diff
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<java version="1.7.0_07" class="java.beans.XMLDecoder">
+<java version="1.7.0_06" class="java.beans.XMLDecoder">
```
akong: Don't check in this file.

Product/checkstyle.settings.xml
```diff
@@ -50,7 +50,6 @@
         <module name="LeftCurly"/>
         <module name="NeedBraces"/>
         <module name="RightCurly"/>
-        <module name="DoubleCheckedLocking"/>
```
akong, line 53: Was this check causing problems?

Product/Production/Services/PatientDiscoveryCore/src/test/java/gov/hhs/fha/nhinc/mpi/adapter/component/hl7parsers/HL7Parser201306Test.java
```diff
@@ -176,8 +184,11 @@ public void testBuildMessageFromMpiPatient_PhoneNumber() {
 
     }
 
+    /**
```
akong, line 187: Empty comment blocks in this class.  Also, out of curiosity, why is checkstyle turned off above?

Product/checkstyle.settings.xml
zmelnick, line 53: Yes. According to the checkstyle documentation for 5.6, the module was removed from checkstyle because the idiom no longer exists in versions of Java 5 and beyond. We are using Java7, so the DoubleCheckedLocking idiom not only does not apply, but it's presence was causing errors for the eclipse checkstyle plugin since it didn't recognize it as a legitimate module.

Product/Production/Services/PatientDiscoveryCore/testMPI.xml
zmelnick: ok

Product/Production/Services/PatientDiscoveryCore/src/test/java/gov/hhs/fha/nhinc/mpi/adapter/component/hl7parsers/HL7Parser201306Test.java
zmelnick, line 187: Ah, I need to actually make a comment for that block. Thanks.

As far as checkstyle being turned off, the class created within this class was causing some errors. Checkstyle was insisting that the class created there needed it's own javadoc, which is fine. But it also wanted the fields to be private and to have public getter and setter methods for each field. It seemed excessive for a test class.

PR #123, opened 2012-10-30
GATEWAY-2902: MDN generation and logging feature.

arathee:
GATEWAY-2902: Image updated with MDN generation and logging feature.


mlandis5am:
## Code Review

General:
- this pull request needs to target the feature-direct branch, not CONNECT_integration.
- it looks like you don't have the latest code from feature-direct.

Product/Production/Common/Properties/Dev/hibernate/hibernate.cfg.xml
- This file can go away as it's replaced by direct.properties file configuration (see a34e76004df4)

...umentSubmission_11/src/main/java/gov/hhs/fha/nhinc/docsubmission/_11/entity/EntityDocSubmissionImpl.java
- This file was refactored - you need to merge in the latest version.
- line 328: This will send multiple MDN messages depending on the number of notification messages in the result... is that we want or is it possible to just send one MDN message response?  If no notification messages exist will we send back an MDN? What is the happy path?  

...ateway/DocumentSubmission_11/src/main/java/gov/hhs/fha/nhinc/docsubmission/_11/entity/MailServer.hbm.xml
- This file can go away as it's replaced by direct.properties file configuration (see a34e76004df4)

...n/Gateway/DocumentSubmission_11/src/main/java/gov/hhs/fha/nhinc/docsubmission/_11/entity/MailServer.java
- Pretty sure this file also goes away with the changes from a34e76004df4.

bhumphrey:
This needs to go to feature_direct

PR #146, opened 2012-11-02
GATEWAY-2947: document query.  add setArguments to builders. web wiring

tparnell:
The integration tests now create BEGIN_INBOUND_MESSAGE events for AdhocQueryRequest.  Next steps are:
- add annotations to all of the methods in the call stack
- modify the description builders to accept all the added params as the stack gets deeper


Product/Production/Services/DocumentQueryCore/src/main/java/gov/hhs/fha/nhinc/docquery/aspect/AdhocQueryResponseDescriptionBuilder.java
```diff
@@ -213,7 +226,10 @@ public boolean apply(ClassificationType type) {
 
     @Override
     public void setArguments(Object... arguments) {
-        // TODO Auto-generated method stub
-
+        if (arguments.length != 1 || !(arguments[0] instanceof AdhocQueryResponse)) {
```
akong, line 229: Do we need to check for null here?

akong:
I'm going to go ahead and pull in this request.

Product/Production/Services/DocumentQueryCore/src/main/java/gov/hhs/fha/nhinc/docquery/aspect/AdhocQueryResponseDescriptionBuilder.java
tparnell, line 229: Will add null check.

PR #148, opened 2012-11-03
Modified aspect advice to use AfterReturning to give event builders acce...

akong:
...ss to the returned object.  Changed modified builders to extractors.


tparnell:
Looks good to me.

PR #149, opened 2012-11-03
Split event processing builders

tparnell:
The @Before and @AfterFinishing invocations may require completely different builders.  Mine do for DQ, for instance.  This adds a new field to the annotation to allow that separation.

### DIFF
--- a/Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/aspect/AdapterDelegationEvent.java
+++ b/Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/aspect/AdapterDelegationEvent.java
@@ -38,13 +38,25 @@ import java.lang.annotation.Target;
 
 /**
  * @author bhumphrey
- *
+ * 
  */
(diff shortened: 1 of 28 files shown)


tparnell:
Forgot that github automatically updates pull requests.  The last commit (384851) is failing integration tests so should not be merged.  But this will give you an idea of the direction I am going.

tparnell:
This morning's commit adds initialization to the assertion extractor.  When I ran the integration tests overnight, they failed with an NPE.  I am re-running the tests now but they likely will not complete before I have to run into the office.

I now expect the integration tests to pass and the request and response objects to be populated for InboundMessage.  I think the other events will also trigger - but am less sure.  More to come.

akong:
The code looks good.   With this, it looks like we're just going to record only the request for @Before and only the response for @AfterReturning rather than trying to have both when available.  That sounds like a good approach.  I'm going to go ahead and pull this request so I can work off of it.

tparnell:
Arthur - one point - you actually WILL have the request objects during @AfterReturning.  I wasn't sure whether we would need those objects to correctly construct the response event.  My DQ impl doesn't use it, but your impl can.

PR #153, opened 2012-11-07
Document Query - gateway builders

tparnell:
Real implementatons of the gateway builders.  Introduces a new transformer builder that delegates after transforming the input and output arguments.


Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/docquery/aspect/ArgTransformerEventDescriptionBuilder.java
```diff
@@ -0,0 +1,181 @@
...
+     * Go from the input return value to the delegate's return value.
+     * 
+     * @param returnValue
+     *            value <code>setReturnValue</code> will be called with
+     * @return value expected by the delegate
+     */
+    protected abstract Object transformReturnValue(Object returnValue);
+
+    private void setMsgObjects() {
+        if (delegate != null) {
```
bhumphrey, line 176: Should this be an assertion? I don't know what the functionality of this should be when the delegate is null.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/docquery/aspect/ArgTransformerEventDescriptionBuilder.java
```diff
@@ -0,0 +1,181 @@
...
+ * then delegates all work to the other builder. Useful for gateway to non-gateway transformations. Concrete
+ * implementations should call <code>setDelegate</code> during construction.
+ */
+public abstract class ArgTransformerEventDescriptionBuilder implements EventDescriptionBuilder {
+    private EventDescriptionBuilder delegate;
+    private MessageRoutingAccessor msgRouting;
+    private EventContextAccessor msgContext;
+
+    @Override
+    public final void buildTimeStamp() {
```
bhumphrey, line 47: I think we need a comment about how this method has become final.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/docquery/aspect/ArgTransformerEventDescriptionBuilder.java
tparnell, line 176: I do the check like this so that external classes can freely mix setDelegate and setMsg\* calls without worrying about a 'correct' ordering.  Everything in this class explodes if delegate is null, so I think it's safe here.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/docquery/aspect/ArgTransformerEventDescriptionBuilder.java
tparnell, line 47: Will add.

PR #156, opened 2012-11-09
GATEWAY-3072: Lightweight gateway implementation for Document Query (Inb...

akong:
...ound only)


Product/Production/Services/DocumentQueryCore/src/main/java/gov/hhs/fha/nhinc/docquery/entity/OutboundDocRetrieveFactory.java
```diff
@@ -0,0 +1,35 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package gov.hhs.fha.nhinc.docquery.entity;
+
+/**
+ * @author akong
+ *
+ */
+public interface OutboundDocRetrieveFactory {
```
akong: This is to be implemented on the next pull request.

Product/Production/Gateway/DocumentQuery_20/src/main/java/gov/hhs/fha/nhinc/docquery/_20/nhin/DocQuery.java
```diff
@@ -26,20 +26,26 @@
...
 @BindingType(value = javax.xml.ws.soap.SOAPBinding.SOAP12HTTP_BINDING)
-@Addressing(enabled=true)
-public class DocQuery implements ihe.iti.xds_b._2007.RespondingGatewayQueryPortType
-{
+@Addressing(enabled = true)
+public class DocQuery implements RespondingGatewayQueryPortType {
+    private InboundDocQueryFactory docQueryFactory;
+
     @Resource
     private WebServiceContext context;
```
bhumphrey, line 50: you will need a setter for this- although it's only required once we aop it.

Product/Production/Gateway/DocumentQuery_20/src/main/resources/applicationContext.xml
```diff
@@ -0,0 +1,44 @@
...
+    xsi:schemaLocation="
+http://www.springframework.org/schema/context http://www.springframework.org/schema/context/spring-context-3.0.xsd
+http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-3.0.xsd
+http://www.springframework.org/schema/aop http://www.springframework.org/schema/aop/spring-aop-3.0.xsd
+">
+    
+    <bean
+        class="org.springframework.beans.factory.config.PropertyPlaceholderConfigurer">
+        <property name="location">
+            <value>file:///#{systemProperties['nhinc.properties.dir']}/gateway.properties</value>
```
bhumphrey: can't this just be file:${nhinc.properties.dir}/gateway.properties  and
should this be gateway.properties or

documentquery.properties

Product/Production/Services/DocumentQueryCore/src/main/java/gov/hhs/fha/nhinc/docquery/inbound/InboundDocQueryFactory.java
```diff
@@ -0,0 +1,36 @@
...
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+package gov.hhs.fha.nhinc.docquery.inbound;
+
+/**
+ * @author akong
+ * 
+ */
+public interface InboundDocQueryFactory {
```
bhumphrey: i don't think we need the factory. the spring context handles that. we could just use InboundDocQuery.

bhumphrey:
The factory classes aren't adding any additional value. With the spring we can inject the instance with out using a factory directly.

Product/Production/Gateway/DocumentQuery_20/src/main/java/gov/hhs/fha/nhinc/docquery/_20/nhin/DocQuery.java
akong, line 50: I purposely did not bring in any aop specific code in.  The setter will be added once the event logging branch gets merged in.

Product/Production/Gateway/DocumentQuery_20/src/main/resources/applicationContext.xml
akong: I believe the file:// prefix is required for spring to see it since it's an external file.  When I left it out, it seems to only looking for it in the ear as a resource.  I will try using ${nhinc.properties.dir} again.  It didn't work for me the first time, but maybe there was something else that was causing it to not get resolved.

I decided on adding it to gateway.properties because in total there will only be one optional property per service.  If that property is not set, then standard is used by default.  But I can add that in if it is better to do so.

Product/Production/Services/DocumentQueryCore/src/main/java/gov/hhs/fha/nhinc/docquery/inbound/InboundDocQueryFactory.java
akong: Ok.  I will remove it.  I added that in as it was in the design wiki approach.

PR #195, opened 2012-11-26
find bugs working - using build tools as plug in dependency

bhumphrey:
The uses build tools as a plugin dependency for checkstyle and findbugs.


tparnell:
BH revising - still 1 minor issue.

PR #213, opened 2012-11-29
GATEWAY-3114 updates to regression suite

sovanncgi:


bhumphrey:
we should also make these executable from maven.

matthewtiller:
Brian, see https://github.com/CONNECT-Solution/CONNECT/pull/202

PR #214, opened 2012-11-29
GATEWAY-3089 complete.

msweaver:


mlandis5am:
Minor Stuff:
Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/InboundMessageHandler.java
- some commented out code on lines 29-30, 142-143

...duction/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/transform/MimeMessageTransformer.java
- you can make transformer a final variable and just new it up in the constructor (or make it a dependency if it needed to be more springified). The accessor method seems superfluous.
- line 25 - this should probably be a real constant - private static final String, uppercase variable name.

...ion/Services/DirectCore/src/test/java/gov/hhs/fha/nhinc/direct/transform/MimeMessageTransformerTest.java
- some line wrapping stuff down around line 90.

PR #224, opened 2012-12-04
Gateway 3183 - admin distribution event logging

tparnell:


akong:
Unit tests for the event annotations?

tparnell:
@akong - I'll add them.  my mistake.

PR #237, opened 2012-12-07
GATEWAY-3191: Fixed the action value when sending a Nhin Request for RD.

akong:


tparnell:
Please add a small unit test for this entire class.

PR #241, opened 2012-12-07
Configs for post connect-a-thon set up

msweaver:
Apparently, I already had a pull request open. This is all my updates. 

Everything is working, with the exception of cert discovery, which cannot be verified due to the port 53 GFE issue.

### DIFF
--- a/Product/Production/Common/Properties/Dev/smtp.agent.config.xml
+++ b/Product/Production/Common/Properties/Dev/smtp.agent.config.xml
@@ -3,32 +3,56 @@
       <AnchorStore type="Uniform" storeType="KeyStore" file="/tmp/directkeystore.jks" filePass="changeit" privKeyPass="changeit"/>    
       <Domain name="direct.connectopensource.org" postmaster="postmaster@direct.connectopensource.org">
          <IncomingTrustAnchors> 
-            <Anchor name="direct.connectopensource.org_ca"/>
-            <Anchor name="direct.connectopensource.org"/>            
-            <Anchor name="direct.securehealthemail.com_ca"/>
+            <Anchor name="infomedtrix"/>
             <Anchor name="direct.careevolutiondirect.com"/>
             <Anchor name="browsersoftdirect.com"/>
+            <Anchor name="direct.connectopensource.org"/>
             <Anchor name="direct.azhisp.com"/>
             <Anchor name="direct.healthvault-ppe.com"/>
             <Anchor name="direct.max.md"/>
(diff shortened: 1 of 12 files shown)


Product/Production/Common/Properties/Dev/smtp.agent.config.xml
```diff
@@ -3,32 +3,56 @@
...
             <Anchor name="ssdev.direct.cert.clinicalinterop.com"/>
             <Anchor name="phimail.emrdirect.com"/>
+            <Anchor name="direct.securehealthemail.com_ca"/>
+            <Anchor name="direct.connectopensource.org_ca"/>
             <Anchor name="direct.icaetc.com"/>
+            <Anchor name="direct.hispdev1.hispdirect.com"/>
+            <Anchor name="common.root.der"/> 
          </OutgoingTrustAnchors>           
       </Domain> 
+      <!--Domain name="5amsolutions.com" postmaster="postmaster@5amsolutions.com">
```
mlandis5am, line 38: I think we can remove this commented out domain.

Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/DirectMailClient.java
```diff
@@ -118,7 +117,9 @@ public void send(MimeMessage message) {
         try {
             MailUtils.sendMessage(message.getAllRecipients(), getMailSession(), message);
         } catch (MessagingException e) {
-            throw new DirectException("Exception while sending mime message.", e);
+        	String errorText = "Exception while sending mime message.";
```
mlandis5am, line 120: I have this logging pushed down into the direct exception constructor, will swap it after this gets pulled in.

Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/MailUtils.java
```diff
@@ -121,8 +121,13 @@ public static void sendMessage(Address[] recipients, Session session, MimeMessag
 
         Transport transport = null;
         try {
-            transport = session.getTransport("smtps");
-            transport.connect();
+            transport = session.getTransport("smtp");
```
mlandis5am, line 124: if this is the "magic sauce" that resolves the mail session connectivity issues, we should rename these "mail.smtp..." properties since they aren't actually javamail properties anymore.

PR #249, opened 2012-12-12
GATEWAY-3244: fixes for processing message multiple times

msweaver:


mlandis5am:
Merging this in so we can collapse the branch.

PR #253, opened 2012-12-15
GATEWAY-3240: Collapsing the feature-direct branch - disabled the direct...

mlandis5am:
... module by default


mlandis5am:
This should fix CI build until we can come up with a better strategy for handling the direct keystores referenced in the smtp.agent.config.xml file.

PR #268, opened 2012-12-20
Gateway-2827: Add targetHomeCommunityId and ServiceName forNhin Service ...

alameluchidambaram:
...call for all Services


Product/Production/Services/DocumentRetrieveCore/src/main/java/gov/hhs/fha/nhinc/docretrieve/nhin/proxy/NhinDocRetrieveProxyWebServiceSecuredImpl.java
```diff
@@ -82,15 +79,20 @@ public RetrieveDocumentSetResponseType respondingGatewayCrossGatewayRetrieve(
                 LOG.debug("After target system URL look up. URL for service: " + sServiceName + " is: " + url);
 
                 if (NullChecker.isNotNullish(url)) {
-                    ServicePortDescriptor<RespondingGatewayRetrievePortType> portDescriptor = getServicePortDescriptor(NhincConstants.GATEWAY_API_LEVEL.LEVEL_g0);
+                    ServicePortDescriptor<RespondingGatewayRetrievePortType> portDescriptor = 
+                            getServicePortDescriptor(NhincConstants.GATEWAY_API_LEVEL.LEVEL_g0);
+                    
+                    String target = null;
+                    if(targetSystem.getHomeCommunity().getHomeCommunityId().startsWith("urn:oid:")) {
+                        target = targetSystem.getHomeCommunity().getHomeCommunityId().replace("urn:oid:", "");
```
akong: Why are we stripping the prefix here and not on the other services?  If this is needed, it should be done in the logic where the target hcid is actually being read.

PR #273, opened 2012-12-20
Gateway-3182:Fix for Evidence Assertion Id

alameluchidambaram:


tparnell:
This change is begging for a unit test.  Can you please create?  It looks like there is history here about the addition or not of the underscore.  I'd like all the cases documented via unit test.

alameluchidambaram:
will add uni test.

alameluchidambaram:
will add unit test.

Product/Production/Common/CONNECTCoreLib/src/test/java/gov/hhs/fha/nhinc/callback/openSAML/OpenSAML2ComponentBuilderTest.java
```diff
@@ -0,0 +1,38 @@
+package gov.hhs.fha.nhinc.callback.openSAML;
```
tparnell, line 29: Missing header.

akong:
An issue that I think needs to be fixed as part of this pull as it deals with incorrect assertion id.  In the class gov.hhs.fha.nhinc.callback.openSAML.HOKSAMLAssertionBuilder, on line 336:

if (evAssertionID == null) {
    evAssertionID = ID_PREFIX.concat(String.valueOf(UUID.randomUUID()));
}

That needs to be fixed as the generated ID must remove all "-" characters like this:

ID_PREFIX.concat(String.valueOf(UUID.randomUUID())).replaceAll("-", "");

Line 81 in the same class does it as such.  So on that class, please create a private method and have both lines 81 and 336 call it instead.

private String createAssertionId() {
    return ID_PREFIX.concat(String.valueOf(UUID.randomUUID())).replaceAll("-", "");
 }

PR #275, opened 2012-12-21
GATEWAY-2957A: Initial pass documenting setup + configuration

mlandis5am:
This is just a README change but I wanted to get this in - there are some placeholders for things that need to be completed like Setting up the public cert in the DNS Zone file.


mlandis5am:
This is just a README change - pulling in.

PR #296, opened 2013-01-03
GATEWAY-3245: AdhocQueryResponse.RegistryObjectList should always be pre...

tparnell:
...sent.


Product/Production/Services/DocRegistryCore/src/main/java/gov/hhs/fha/nhinc/docregistry/adapter/proxy/AdapterComponentDocRegistryProxyNoOpImpl.java
```diff
@@ -42,6 +43,8 @@
 
     public AdhocQueryResponse registryStoredQuery(AdhocQueryRequest request, AssertionType assertion) {
         LOG.trace("Using NoOp Implementation for Adapter Component Doc Registry Service");
-        return new AdhocQueryResponse();
+        AdhocQueryResponse response = new AdhocQueryResponse();
```
akong, line 47: Do we need a utility method to create this response object?  Since it seems like we shouldn't be creating one without an RegistryObjectList (and maybe status?), it might make sense to just wrap it.

Product/Production/Services/DocumentQueryCore/src/main/java/gov/hhs/fha/nhinc/docquery/adapter/proxy/AdapterHelper.java
```diff
@@ -0,0 +1,52 @@
...
+import oasis.names.tc.ebxml_regrep.xsd.query._3.AdhocQueryResponse;
+import oasis.names.tc.ebxml_regrep.xsd.rim._3.RegistryObjectListType;
+import oasis.names.tc.ebxml_regrep.xsd.rs._3.RegistryError;
+import oasis.names.tc.ebxml_regrep.xsd.rs._3.RegistryErrorList;
+
+class AdapterHelper {
+
+    public AdhocQueryResponse createErrorResponse() {
+        AdhocQueryResponse response;
+        response = new AdhocQueryResponse();
```
akong: Combine the two lines above.

Product/Production/Services/PatientDiscoveryCore/testMPI.xml
```diff
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<java version="1.7.0_06" class="java.beans.XMLDecoder">
```
akong, line 2: Don't check in.

Product/Production/Adapters/DocumentQuery_a0/src/test/java/gov/hhs/fha/nhinc/docquery/adapter/AdapterDocQueryOrchImplTest.java
```diff
@@ -45,6 +45,14 @@ protected AdapterRedactionEngineProxy getRedactionEngineProxy() {
         }
     }
 
+    @Test
+    public void errorResponseHasRegistryObjectList() {
+        AdapterDocQueryOrchImpl impl = new AdapterDocQueryOrchImpl();
+        AdhocQueryResponse response = impl.respondingGatewayCrossGatewayQuery(null, null);
+        assertNotNull(response.getRegistryObjectList());
+        assertEquals(0, response.getRegistryObjectList().getIdentifiable().size());
```
tparnell, line 53: Not using the full validator here on purpose.  This helper method only has responsibility for partial construction.

Product/Production/Services/DocRegistryCore/src/main/java/gov/hhs/fha/nhinc/docregistry/adapter/proxy/AdapterComponentDocRegistryProxyNoOpImpl.java
tparnell, line 47: I decided against it.  There are many variables - the helper would end up being really if/elseif/elseif

PR #299, opened 2013-01-04
GATEWAY-3189: Removed unused properties from corelib.

akong:
All file deletions with one file move (truststore.properties).  Please see ticket GATEWAY-3189 for details.


Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/docrepository/adapter/util/DocumentLoadUtil.java
```diff
@@ -1,473 +0,0 @@
```
msweaver: I don't have a good reason to keep this and the next java file, but I am curious about the rational for deleting them? I am assuming they are unused...

msweaver:
Other than a quick discussion on the comment I posted I am ready to merge this. Maybe we can skype for 2 seconds in the morning?

PR #304, opened 2013-01-07
GATEWAY-3306: Read security properties file from nhinc properties.

akong:


tparnell:
The code you have so far looks great.  A couple things seems to be missing: (1) the new configuration files, and (2) documentation on where the new config is.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/messaging/service/decorator/cxf/WsSecurityConfigFactory.java
```diff
@@ -0,0 +1,122 @@
...
+        return INSTANCE;
+    }
+
+    /**
+     * Returns a cloned copy of the map containing all the necessary security configuration. This includes all the
+     * properties read from the signature.properties file.
+     * 
+     * @return
+     */
+    public Map<String, Object> getConfiguration() {
```
tparnell, line 81: Consider returning ImmutableMap from google collections so you don't have to deep copy each time.

Product/Production/Common/CONNECTCoreLib/src/test/java/gov/hhs/fha/nhinc/messaging/service/decorator/cxf/WsSecurityConfigFactoryTest.java
```diff
@@ -0,0 +1,103 @@
+/*
+ * Copyright (c) 2012, United States Government, as represented by the Secretary of Health and Human Services.
```
tparnell, line 2: Please update your local eclipse settings to the new license text.

akong:
There weren't any new config files added as part of this issue.  The only change done here was renaming keystore.properties to signature.properties.  The config files already exists in the Properties/Dev directory (saml.properties, truststore.properties, and signature.properties).  On a build+deploy, these files are jar-ed into a Properties.jar and also exploded into config/nhin directory.  Previously, these files are read from the Properties.jar.  This change will make the code look for it in the config/nhin directory instead.

Product/Production/Common/CONNECTCoreLib/src/test/java/gov/hhs/fha/nhinc/messaging/service/decorator/cxf/WsSecurityConfigFactoryTest.java
akong, line 2: Will fix.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/messaging/service/decorator/cxf/WsSecurityConfigFactory.java
akong, line 81: Using the ImmutableMap was my first cut at this code.  But for some reason, ws-security was failing during integration when I did that.  I'm not sure if it is because cxf needs to modify the map or if it is because I had to make the Properties instance (cryptoProperties) inside the config map into an ImmutableMap as well, but it wasn't working so I decided to just use deep copy.

PR #305, opened 2013-01-08
Gateway 3278

tparnell:


Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/xmlCommon/XmlUtility.java
```diff
@@ -147,20 +147,6 @@ public static Node performXpathQuery(String sourceXml, String xpathQuery) throws
...
     /**
      * @deprecated Use {@link XpathHelper} instead
-     * @param sourceXml
-     * @param xpathQuery
-     * @param namespaceContext
-     * @return
-     * @throws javax.xml.xpath.XPathExpressionException
-     */
-    @Deprecated
-    public static Node performXpathQuery(String sourceXml, String xpathQuery, NamespaceContext namespaceContext)
```
tparnell, line 157: Only a test case was calling this method.  So all the removals that use namespacecontext don't affect production code.

PR #307, opened 2013-01-09
GATEWAY-3321: make connect have only one set of hibernate files

bhumphrey:
consolidating the hibernate config files to use the full jndi names by prefixing them with
java:comp/env/

[GATEWAY-3321](https://issues.connectopensource.org/browse/GATEWAY-3321)


tparnell:
Exact same comment as weaver: if you give a thumbs up for weblogic, this is immediately mergable.

PR #312, opened 2013-01-09
Regression suite updates

jsmith2:
Updated all regression suites with minor changes for recent updates including:
- Refreshing interfaces
- Adding correct assertion values to event logging and error codes.
- Updating incorrect proxy namespaces to entity.
- Deleting any proxy testing info, test cases.


jsmith2:
You can ignore all the updated date garbage.

PR #316, opened 2013-01-12
GATEWAY-3273: Added unit tests for core mail package.

mlandis5am:
- unit tests for core mail package.
- reworked poms to inherit versions from dependency-management


Product/Production/Gateway/AdminDistribution_10/pom.xml
```diff
@@ -49,6 +49,10 @@
```
msweaver, line 48: why is this necessary? Same comment for all of these Gateway project poms.

msweaver:
As a general comment, having stepped back from Direct for a little bit and now coming back to see it again, I am second guessing the placement of this logic into CONNECTCoreLib. Only DirectCore uses this functionality, it seems like all of this would be at home in DirectCore. 

There aren't any prospects of integrating other services into Direct right?

Product/Production/Gateway/AdminDistribution_10/pom.xml
mlandis5am, line 48: Either the compile or runtime tests failed without the explicit dependency (I just can't remember which at this moment). I modified the Product/pom.xml below to specify the scope "test" for the spring test jars and after that the jar went missing in these modules without these entries added. AFAIK this is the right way to use maven but I could be missing something here.

mlandis5am:
I updated this to remove the nhincdb.sql changes - they will come separately with 3222.

PR #317, opened 2013-01-14
Gtaeway-3283:Fix Properties file having same names

alameluchidambaram:
The Properties filname has been changed those which are having same names for different services like DocumetRetrieve_20 and DocumentRetrieve_30 to resolve websphere classloader problems.


msweaver:
I know this sounds nit-picky but getting gateway spelled right is kind of important because some one can come by later and search for "gateway" and then not find this commit. Checkout this link:
http://stackoverflow.com/questions/179123/how-do-i-edit-an-incorrect-commit-message-in-git

PR #323, opened 2013-01-16
GATEWAY-3314: DQ Fan-out fix.

akong:


Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/messaging/service/port/CachingCXFServicePortBuilder.java
```diff
@@ -3,35 +3,77 @@
...
-        T port = (T) CACHED_PORTS.get(serviceEndpointClass);
+    public synchronized T createPort() {
+        T port = (T) getCache().get(serviceEndpointClass);
         if (port == null) {
-                   port  = super.createPort();
-            ((BindingProvider) port).getRequestContext().put(
-                    "thread.local.request.context", Boolean.TRUE); 
-            CACHED_PORTS.put(serviceEndpointClass, port);
+            port = super.createPort();
+            ((BindingProvider) port).getRequestContext().put("thread.local.request.context", Boolean.TRUE);
```
tparnell: Extract & document where this came from.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/transform/marshallers/Marshaller.java
```diff
@@ -75,6 +80,42 @@ public Element marshal(Object object, String contextPath) {
...
+    /**
+     * Marshalls the passed in XML Bean object into an XML Element.  
+     * 
+     * @param object - the XML bean to be marshalled
+     * @param contextPath - The name of the context. (i.e. "org.hl7.v3").
+     * @param qname - the qualified name of the XML (i.e. "urn:org:hl7:v3")
+     * @return the XML Element
+     */
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    public Element marshal(Object object, String contextPath, QName qname) {
```
tparnell, line 92: Does performance matter here?  I believe it's quite expensive to create the context.  (or one of the objects at least).  And it's cheap to do the marshalling.  Recommend a caching mechanism or at least a note that this shouldn't be used in performance-critical sections.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/util/MessageGeneratorUtils.java
```diff
@@ -37,6 +45,12 @@
 
     private static MessageGeneratorUtils INSTANCE = new MessageGeneratorUtils();
 
+    private static String NHINC_COMMON_CONTEXT = "gov.hhs.fha.nhinc.common.nhinccommon";
+    private static String NHINC_COMMON_URN = "urn:gov:hhs:fha:nhinc:common:nhinccommon";
```
tparnell, line 49: prefer "urn:" + NHINC_COMMON_CONTEXT;

Product/Production/Services/DocumentQueryCore/src/main/java/gov/hhs/fha/nhinc/docquery/entity/AggregationService.java
```diff
@@ -229,51 +234,70 @@ protected AdhocQueryRequest setTargetHomeCommunityId(AdhocQueryRequest request,
...
+    
+    protected AssertionType cloneAssertionWithNewMsgId(AssertionType assertion) {
+        AssertionType newAssertion = MessageGeneratorUtils.getInstance().clone(assertion);
+        
+        newAssertion.setMessageId(new WSAHeaderHelper().generateMessageID());
+        
+        return newAssertion;
+    }
+    
+    protected Set<II> removeDuplicates(List<II> iiArray) {
```
tparnell, line 248: This feels like a dangerous implementation.  It is possible (though unlikely) that external callers are relying on the default equality/hashcode behavior.  Silently substituting a subclass doesn't seem to fit what you are trying to accomplish here.  Perhaps you can have ComparableII keep the original II around as a private field, then transform the final set back into the original IIs before returning?

Product/Production/Services/DocumentQueryCore/src/main/java/gov/hhs/fha/nhinc/docquery/entity/OutboundDocQueryDelegate.java
```diff
@@ -93,7 +93,7 @@ public OutboundDocQueryOrchestratable process(OutboundDocQueryOrchestratable mes
 
         if (response instanceof OutboundDocQueryOrchestratable_a0) {
             LOG.debug("NhinDocQueryDelegate::process returning a0 result");
-        } else if (response instanceof OutboundDocQueryOrchestratable_a0) {
+        } else if (response instanceof OutboundDocQueryOrchestratable_a1) {
```
tparnell, line 96: How did you discover this?  This change seems unrelated to the rest of the commit.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/transform/marshallers/Marshaller.java
akong, line 92: The JAXBContext are being cached by the JAXBContextHandler, but yes, there is some performance hit.   But I felt it was worth it to not write our own clone method which will need to change if the schema changes.  A better approach might be to use a xjc plugin that will insert a clone() implementation on the generated sources, but I did not want to mess around with the Schema generation plugin this late in the game.

I updated a ticket that is related to this:

https://issues.apache.org/jira/browse/CXF-3354

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/util/MessageGeneratorUtils.java
akong, line 49: The urn is actually different as it is delimited by ":" rather than ".".

Product/Production/Services/DocumentQueryCore/src/main/java/gov/hhs/fha/nhinc/docquery/entity/OutboundDocQueryDelegate.java
akong, line 96: This was causing an invalid log statement and confused me when I was debugging DQ fan out as it was logging something that wasn't happening.  This if-else statement does nothing but log so it's only a minor change...

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/messaging/service/port/CachingCXFServicePortBuilder.java
akong: Will do.

Product/Production/Services/DocumentQueryCore/src/main/java/gov/hhs/fha/nhinc/docquery/entity/AggregationService.java
akong, line 248: Ok.  I'll fix it.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/util/MessageGeneratorUtils.java
tparnell, line 49: Ah, yes.  My eyes missed that.

PR #326, opened 2013-01-16
GATEWAY-3351: code updated and unit tests fixed.

msweaver:


Product/Production/Services/DocumentQueryCore/src/main/java/gov/hhs/fha/nhinc/docquery/outbound/PassthroughOutboundDocQuery.java
```diff
@@ -77,23 +74,19 @@ public AdhocQueryResponse respondingGatewayCrossGatewayQuery(AdhocQueryRequest r
 
         String targetHCID = getTargetHCID(target);
 
-        auditRequestFromAdapter(request, assertion, targetHCID);
```
akong, line 80: Passthrough does not require audits from the adapter?

Product/Production/Services/DocumentQueryCore/src/main/java/gov/hhs/fha/nhinc/docquery/outbound/PassthroughOutboundDocQuery.java
```diff
@@ -77,23 +74,19 @@ public AdhocQueryResponse respondingGatewayCrossGatewayQuery(AdhocQueryRequest r
...
-
         return response;
     }
 
-    private String getTargetHCID(NhinTargetSystemType target) {
-        String targetHCID = null;
-        if (target != null && target.getHomeCommunity() != null) {
-            targetHCID = target.getHomeCommunity().getHomeCommunityId();
-        }
+	private String getTargetHCID(NhinTargetSystemType target) {
```
akong: The spacing/code formatting looks weird here and the other files...

msweaver: I guess I missed the ol Command+Shift+F on this file.

Product/Production/Services/DocumentQueryCore/src/main/java/gov/hhs/fha/nhinc/docquery/outbound/PassthroughOutboundDocQuery.java
msweaver, line 80: Passthru assumes the adapter does pretty much everything except SAML, and uddi look ups.

akong, line 80: In this case, passthru is still doing audits for outbound messages to the nhin since the delegate does audits as well (i.e. with this change, passthru has 2 audits outbound instead of 4).  We've been inconsistent with it on previous releases, but with LWGW, passthru is auditing (or supposed to) the same way as standard (4 on the requesting and 4 on the responding) with the idea that people can turn it off in the spring proxy config file.

msweaver, line 80: Im not sure about that approach, there is still a lot of processing that goes on to format a message before it’s sent to auditing. In terms of the message audited to the Nhin, I believe that was always audited.

akong, line 80: OK.  Per discussion, GATEWAY-3180 will address what to do with auditing and Passthrough for other services.

PR #337, opened 2013-01-21
GATEWAY-3377: Fixed library conflicts and missing db resource for Univer...

akong:
...sal Client GUI.

The following was modified to get Universal Client GUI to work properly.
1. Added db resource in web.xml
2. Removed test scope for cglib
3. Forced UClient to use jsf-extensions library that are in the ThirdParty/local repo.  Previously, we were mixing jars from the local repo and remote repo.  Apparently, the remote jsf-extensions jars were not compatible with the other dependent libs from the local repo, and since the correct ones were already in our local repo and doesn't seem to exists in the remote repo, I changed the UClient pom.xml file to just use everything from local.


akong:
Closing pull to wait for Todd's fixes.

PR #344, opened 2013-01-23
Gateway 3382

alameluchidambaram:
Fix Audit Logging for Document Retrieve Service


Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/util/HomeCommunityMap.java
```diff
@@ -144,15 +143,19 @@ public static String getCommunityIdFromAssertion(AssertionType assertion) {
...
             }
 
-        } else if (assertion != null && assertion.getHomeCommunity() != null) {
+        }
+        if (communityId == null && assertion != null && assertion.getHomeCommunity() != null) {
             communityId = assertion.getHomeCommunity().getHomeCommunityId();
         }
+        if (NullChecker.isNullish(communityId)){
+            LOG.error("CommunityId in Assertion passed is Null");
+        }
```
jsmith2, line 154: If you know this is null at this point, you should just return the communityID, no need passing it to the formatHomeCommunityId.

PR #355, opened 2013-01-24
Gateway-3387:Fix Audit Logging for DR Passthru Service

alameluchidambaram:


bhumphrey:
I don't see any tests that verify that this is correct.

alameluchidambaram:
I will do the unit test to verify it

PR #357, opened 2013-01-24
GATEWAY-3388: Fixed DS passthru audit.

akong:
For inbound:

Moved the sendToAdapter() code from passthrough to the abstract parent class.  Removed the audits in that method to make passthrough not audit.  Added audits in the standard and made it call sendToAdapter()

For outbound:
Removed the audits from the passthrough code.

### DIFF
--- a/Product/Production/Services/DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/docsubmission/inbound/StandardInboundDocSubmission.java
+++ b/Product/Production/Services/DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/docsubmission/inbound/StandardInboundDocSubmission.java
@@ -32,6 +32,7 @@ import gov.hhs.fha.nhinc.common.nhinccommon.AssertionType;
 import gov.hhs.fha.nhinc.docsubmission.MessageGeneratorUtils;
 import gov.hhs.fha.nhinc.docsubmission.XDRAuditLogger;
 import gov.hhs.fha.nhinc.docsubmission.XDRPolicyChecker;
+import gov.hhs.fha.nhinc.docsubmission.adapter.proxy.AdapterDocSubmissionProxyObjectFactory;
 import gov.hhs.fha.nhinc.nhinclib.NhincConstants;
 import gov.hhs.fha.nhinc.nhinclib.NullChecker;
 import gov.hhs.fha.nhinc.properties.PropertyAccessException;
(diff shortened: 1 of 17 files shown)


Product/Production/Services/DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/docsubmission/inbound/StandardInboundDocSubmission.java
```diff
@@ -32,6 +32,7 @@
 import gov.hhs.fha.nhinc.docsubmission.MessageGeneratorUtils;
```
msweaver, line 32: I think these changes, and the changes around the AbstractInbound.. class, as well as the subsequent changes to Passthru belong in their own commit. 

These changes really really a fix for standard mode.

Product/Production/Services/DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/docsubmission/inbound/deferred/request/StandardInboundDocSubmissionDeferredRequest.java
```diff
@@ -31,6 +31,7 @@
```
msweaver, line 30: same comment.

Product/Production/Services/DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/docsubmission/inbound/deferred/response/StandardInboundDocSubmissionDeferredResponse.java
```diff
@@ -26,12 +26,11 @@
```
msweaver, line 25: same comment here again.

Product/Production/Services/DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/docsubmission/inbound/StandardInboundDocSubmission.java
akong, line 32: As in a different commit/different ticket?  I think they are all related as the decoupling from passthrough is needed so that I can remove the audits from passthrough and still audit in standard.  Otherwise the changes in passthrough will break existing functionality.

As in, before this ticket, standard is auditing properly.  If possible, I don't want one commit to break that and another to fix it.

akong, line 32: Per discussion, agreed that although it would have been ideal to do the decoupling in GATEWAY-3180 and its DS subtask, it also makes sense doing it in here as this ticket is not committable without the changes required to standard (i.e. will break existing functionality).

PR #363, opened 2013-01-25
Gateway-3411 Updated Install for configurable deploy version

jsmith2:
- Pulled deployment.version out of build.xml and into install.properties.  Makes it configurable for installing different versions of Connect ear.
- Anytime a new tag is created or the repository name is changes, the install.properties should be updated for the tag/update.

### DIFF
--- a/Product/Install/install.properties
+++ b/Product/Install/install.properties
@@ -14,4 +14,3 @@ direct.anchor.store=anchors
 direct.trust.store=example-store
 direct.key.store=example-key
 skip.backup=false
-deployment.version=4.0.0-SNAPSHOT
--- a/Product/Install/install.properties
+++ b/Product/Install/install.properties
@@ -13,3 +13,5 @@ cert.store.name=cacerts
 direct.anchor.store=anchors
 direct.trust.store=example-store
 direct.key.store=example-key
+skip.backup=false
+deployment.version=4.0.0-SNAPSHOT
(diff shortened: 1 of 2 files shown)


tparnell:
If we go this route, every time we rev the maven poms, we need to manually update the properties file.  This type of double update violates DRY and is error prone.  Please look at Product/build.xml:51-53 to see how to pull the current pom version automatically.  My recommendation is as follows:
- default to the version from the pom (via the xml extraction), and
- allow override via install.properties or the command line

bhumphrey:
Agree with Todd on this. I think it should be bound to the version that is was built for. Like it is in the Product/build.xml.

msweaver:
For reference:

```
    <xmltask source="pom.xml" >
       <copy path="//*[local-name() = 'version']/text()" property="maven_version"/>
    </xmltask>
```

Then use the maven_version property like any other property (ie get the value with ${maven_version} ).

jsmith2:
Will check on the Product/build.xml example.  Thanks.

Product/Install/install.properties
```diff
@@ -13,3 +13,5 @@ cert.store.name=cacerts
 direct.anchor.store=anchors
 direct.trust.store=example-store
 direct.key.store=example-key
+skip.backup=false
+deployment.version=4.0.0-SNAPSHOT
```
bhumphrey: is this still necessary?

PR #366, opened 2013-01-30
Gateway 3390

msweaver:
Subtask GATEWAY-3423, updates for creating a websphere enterprise EAR.


Product/Production/Deploy/pom.xml
```diff
@@ -32,6 +32,14 @@
             </modules>
         </profile>
         <profile>
+            <id>webspherece</id>
```
mlandis5am: IMHO the new profile should be for WAS - not for WASCE. Otherwise, this has a downstream effect on all of our WASCE adopters and our documentation.

Product/Production/Deploy/websphere/pom.xml
```diff
@@ -7,18 +7,17 @@
```
mlandis5am, line 6: Similar to last comment: I think this pom should be for WASCE and the new pom for WAS.

msweaver, line 6: It’s a two line documentation change, and this naming more accurately reflects reality. In the wild the term “WebSphere” refers to WAS Enterprise, not WASCE.

mlandis5am, line 6: So... The bigger concern is that it also impacts adopters actually using the profile.  The ..-WS.ear now means something different, but the ear will still be produced by the profile and adopters will try to deploy it -- which will likely result in mysterious errors.

msweaver, line 6: The -WS ear doesn't exist anymore, there is a CONNECT-WAS.ear and a CONNECT-WASCE.ear.

PR #374, opened 2013-02-04
GATEWAY-3445: changes for hp uddi to not use mtom.

msweaver:


Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/messaging/client/CONNECTCXFClientUnsecuredNoMTOM.java
```diff
@@ -0,0 +1,68 @@
+/*
+ * Copyright (c) 2012, United States Government, as represented by the Secretary of Health and Human Services.
```
tparnell, line 2: Wrong copyright.  Please update eclipse settings.

tparnell:
Because UDDI doesn't need the Connect business logic associated with id matching, and because we don't want to have the same configuration logic in two classes, closing without merge.  BH will be submitting his patch for review.

PR #383, opened 2013-02-05
added HTTPClientPolicyHelper 

bhumphrey:
This is to remove dependency between ServicePortBuilder and EndpointDecorators.


Product/Production/Common/CONNECTCoreLib/src/test/java/gov/hhs/fha/nhinc/messaging/service/ServiceEndpointTest.java
```diff
@@ -0,0 +1,110 @@
...
+        
+        assertTrue(requestContext.containsKey(HTTPClientPolicy.class.getName()));
+        
+        
+    }
+    
+    
+    @Test
+    public void verifyExsistingOneIsUsedWhenPresent() {
+        BindingProvider bindingProvider = mock(BindingProvider.class);
```
tparnell: Repeated lines of code between this method and the one above it.

Product/Production/Common/CONNECTCoreLib/src/test/java/gov/hhs/fha/nhinc/messaging/service/ServiceEndpointTest.java
```diff
@@ -0,0 +1,110 @@
...
+ */
+public class ServiceEndpointTest {
+
+    @Test
+    public void verifyNewIsCreatedWhenNull() {
+        BindingProvider bindingProvider = mock(BindingProvider.class);
+        Map<String, Object> requestContext = new HashMap<String, Object>();
+        
+        when(bindingProvider.getRequestContext()).thenReturn(requestContext);
+        assertFalse(requestContext.containsKey(HTTPClientPolicy.class.getName()));
```
tparnell: This isn't asserting on the system under test.  You appear to just checking that an empty hashmap is indeed empty.  This isn't needed.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/messaging/service/HTTPClientPolicyHelper.java
```diff
@@ -0,0 +1,62 @@
...
+public class HTTPClientPolicyHelper {
+
+    
+    /**
+     * Get the {@link HTTPClientPolicy} from the request context of bindingProvider, creates the  {@link HTTPClientPolicy} if null.
+     * 
+     * @param bindingProvider provider to get the {@link HTTPClientPolicy} from
+     * @return the existing {@link HTTPClientPolicy} or a new one.
+     */
+     static HTTPClientPolicy getHTTPClientPolicy(BindingProvider bindingProvider) {
```
tparnell: Prefer non-static to static, so this class can be mocked/injected in the future.

PR #405, opened 2013-02-19
Problem with build with removal of Dev folder

jsmith2:
Dev folder was previously removed but the folder is still referenced in some of our build files.


bhumphrey:
add to whitelist

PR #409, opened 2013-02-21
remove performance repository

bhumphrey:
There were no test for this code and it is deprecated (replaced by the event logging)


jsmith2:
Need to remove the perfRepo from Product/DBScripts in nhincdb/nhincdb.sql, nhincdb/dropall.sql, perfrepo directory, and the oracle db scripts as well.

jsmith2:
And the domain.xml (self signed) needs to be modified to remove the data source.

bhumphrey:
There is another issue in jira for this. This pull is just to help code
coverage.

Sent from my iPhone

On Feb 24, 2013, at 11:28 AM, Jason notifications@github.com wrote:

And the domain.xml (self signed) needs to be modified to remove the data
source.

—
Reply to this email directly or view it on
GitHubhttps://github.com/CONNECT-Solution/CONNECT/pull/409#issuecomment-14010992.

jsmith2:
I added a comment to that issue (https://issues.connectopensource.org/browse/CONN-84), that the code removal part was complete.

PR #438, opened 2013-03-14
CONN-40:Fix PD Exceptions in log

alameluchidambaram:


Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/transform/audit/PatientDiscoveryTransforms.java
```diff
@@ -1146,7 +1146,7 @@ protected String getHCIDFromReceiver(PRPAIN201305UV02 requestMessage) {
             communityId = requestMessage.getReceiver().get(0).getDevice().getAsAgent().getValue()
                 .getRepresentedOrganization().getValue().getId().get(0).getRoot();
         } catch (NullPointerException | IndexOutOfBoundsException e) {
-            LOG.debug("could not obtain HCID from Receiver RepresentedOrganization.", e);
+            LOG.debug("could not obtain HCID from Receiver RepresentedOrganization.");
```
bhumphrey, line 1149: I think the exception should be logged but at the error level instead of the debug level.

msweaver, line 1149: I think these changes reflect what's being asked for in the ticket. The issue is that an exception is being reported when the hcid isn't found, but the hcid is being search for in two places. Therefore an alarming exception is being displayed when there really isn't a problem.

PR #439, opened 2013-03-14
Conn 183

msweaver:
I messed up pulling in the changes from CONNECT_integration on a previous pull that I merged (so the 4.1.0-SNAPSHOT got reintroduced when 4.0.1-SNAPSHOT is correct), here is the fix for that along with 183.


Product/Production/Services/DirectCore/src/test/java/gov/hhs/fha/nhinc/direct/DirectAdapterTest.java
```diff
@@ -51,7 +51,11 @@
```
bhumphrey, line 50: seems un-related to version changes

Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/DirectAdapter.java
```diff
@@ -31,7 +31,6 @@
```
bhumphrey, line 30: seems un-related to version changes

PR #441, opened 2013-03-14
CONN-92 Remove NHINC_PROPERTIES_DIR Env Variable

jsmith2:
- Removed all references to NHINC_PROPERTIES_DIR
- validation_testing_driver.xml no longer used, so removed.
- PropertyAccessorFileUtilities was formatted, so whole file appears changed, only review the checkPropertyFileDir() for changes.


Product/Production/Services/HIEMCore/src/main/java/gov/hhs/fha/nhinc/hiem/configuration/topicconfiguration/TopicConfigurationManager.java
```diff
@@ -161,7 +160,6 @@ private static void loadTopicConfigurations() throws ConfigurationException {
         topicConfigurations = TopicConfigurationsXML.deserialize(sTopicConfigXml);
 
         if (topicConfigurations != null) {
-            m_bInternalLoaded = true;
```
msweaver, line 164: This seems weird. Maybe change the if statement to but == null, add the log statement to the if part and remove the else?

Product/Install/validation_testing_driver.xml
```diff
@@ -1,446 +0,0 @@
```
msweaver: Why delete this file?

jsmith2: This was used for the old CI pipeline and has been subsequently replaced by maven.

msweaver: ok makes sense.

Product/Production/Services/HIEMCore/src/main/java/gov/hhs/fha/nhinc/hiem/configuration/topicconfiguration/TopicConfigurationManager.java
jsmith2, line 164: This was an unintended consequence of removing a class variable marked as unused.  I'll update that to be a if(== null) and get rid of the else.

PR #462, opened 2013-04-05
Conn 355:Conformance for AdapterPolicyEngineSAMLJavaImpl

alameluchidambaram:


Product/Production/Services/PolicyEngineCore/src/main/java/gov/hhs/fha/nhinc/policyengine/adapter/orchestrator/proxy/AdapterPolicyEngineOrchProxyJavaImpl.java
```diff
@@ -29,9 +29,16 @@
```
msweaver, line 28: This file should be unchanged. I think you need to back these changes out.

PR #489, opened 2013-06-06
CONN-376: Regression suite for 376 (which also tests 418).

msweaver:


nareshcgi:
This url needs to be fixed in CONN-376 / CONN-376-soapui-project.xml, SOAPUI is not able to find the location. 
definition="../../../NHINC/Interfaces/wsdl/AdapterDocQuery.wsdl"

PR #497, opened 2013-06-12
CONN-440: Fixed code to populate WSA:To value

nareshcgi:


Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/messaging/client/CONNECTCXFClientFactory.java
```diff
@@ -48,26 +48,33 @@
...
      * passed for HIEM use.
      */
+    @Override
     public <T> CONNECTClient<T> getCONNECTClientSecured(ServicePortDescriptor<T> portDescriptor, String url,
-            AssertionType assertion, String wsAddressingTo, String subscriptionId) {
-        return new CONNECTCXFClientSecured<T>(portDescriptor, url, assertion, wsAddressingTo, subscriptionId);
+        AssertionType assertion, String wsAddressingTo, String subscriptionId) {
+        String wsAddressingToValue = wsAddressingTo;
+        //use the url if the wsaddressing is null
+        if (wsAddressingToValue == null) {
```
msweaver: What if the url is whitespace or empty but not null? Shouldn't we use StringUtils.isBlank?

PR #499, opened 2013-06-13
CONN-469:Create RealtesToList for all Services in RegressionSuite

alameluchidambaram:


nareshcgi:
Change the Manual Step description and expected result comments...give the exact location where to look for the RelatesTo element.

PR #503, opened 2013-06-18
CONN-472:Add content-type check for PD and DQ Services

alameluchidambaram:


nareshcgi:
SOAPUI is not able to find the wsdl location for DocumentQuery and PatientDiscovery projects.

PR #524, opened 2013-07-20
CONN-517:Fix DQ for fan-out suite

alameluchidambaram:


Product/Production/Gateway/DocumentQuery_20/src/main/java/gov/hhs/fha/nhinc/docquery/_20/entity/EntityDocQueryImpl.java
```diff
@@ -95,6 +95,9 @@ private AdhocQueryResponse respondingGatewayCrossGatewayQuery(AdhocQueryRequest
         try {
             if (targets != null && StringUtils.isBlank(targets.getUseSpecVersion())) {
                 targets.setUseSpecVersion("2.0");
+            } else {
+                targets = new NhinTargetCommunitiesType();
```
msweaver, line 101: This is incorrect and will introduce a bug. If the targets.getUseSpecVersion() is not blank (move to the else condition) it's already been populated in the entity message, and the current logic will override it.

Do this instead. Break up the first if condition. If targets is null then go ahead and create the targets object. After that check if use spec version is blank.

PR #544, opened 2013-08-16
CONN-490:JBOSS CI Scripts

alameluchidambaram:


Product/Install/JBoss/JBOSS_Variables.sh
```diff
@@ -0,0 +1 @@
+export EARFILENAME=`find  /home/jenkins/workspace/CONNECT_CI_LINUX_JBOSS/.repository/org/connectopensource/CONNECT-JB7/ -name "*.ear"`
```
msweaver, line 1: How do we know this won't find multiple instances of the jboss ear from different versions?

alameluchidambaram, line 1: There is Only one ear exists at any point because we clean up everything in /workspace/CONNECT_CI_LINUX_JBOSS/.repository/ before we start extracting JBoss ear from Glassfish Parent job

Product/Install/JBoss/deployApp.sh
```diff
@@ -0,0 +1,13 @@
+#Script to Deploy CONNECT APP to JBoss 7.1.1Final Scripts
```
msweaver, line 1: Are we missing a start server command?

alameluchidambaram, line 1: There is a start scipt separately startServer.sh

msweaver, line 1: shouldn't it be called from here? it seems like we are waiting for the start up out of the blue.

alameluchidambaram, line 1: It's called in Jenkins this way startServer script first and then deployApp.

The deployApp will wait for the server before it deploys.

/nhin/CI/startServer.sh

/nhin/CI/deployApp.sh

PR #545, opened 2013-08-19
Some README data does not display correctly on the GitHub website

zmelnick:
Specifically, the values of <Soapui Home> does not appear in the testing section, leading users to believe that the additional jars need to be put in the root /bin directory. This is probably a result of .md markup. The <> notation has been changed so that the values appear correctly when viewed on the GitHub site.

Also, entity and msgproxy validation files no longer exist. They have been replaced with a single connect validation concept.


msweaver:
Ok to test

PR #554, opened 2013-08-23
Conn 535

nareshcgi:


Product/Production/Services/DocumentRetrieveCore/src/main/java/gov/hhs/fha/nhinc/docretrieve/entity/OutboundDocRetrieveStrategyBase.java
```diff
@@ -44,8 +45,15 @@ public void execute(OutboundDocRetrieveOrchestratable message) {
 
         if (message instanceof OutboundDocRetrieveOrchestratable) {
             OutboundDocRetrieveOrchestratable NhinDRMessage = (OutboundDocRetrieveOrchestratable) message;
+            //Append urn:oid to the home community id if its not present
             String requestCommunityID = HomeCommunityMap.getCommunityIdForRDRequest(NhinDRMessage.getRequest());
 
+            //Assign the modified request community id value to the request
+            if (NhinDRMessage.getRequest() != null && NullChecker.isNotNullish(NhinDRMessage.getRequest().getDocumentRequest())
```
msweaver: I don't think this is right. Why doesn't it iterate through all of the document requests?

PR #563, opened 2013-09-05
Conn 575

alameluchidambaram:


Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/configuration/jmx/PassthruMXBeanRegistry.java
```diff
@@ -50,7 +50,7 @@
     }
 
     /** The registered beans. */
-    Set<WebServicesMXBean> registeredBeans = null;
+    public Set<WebServicesMXBean> registeredBeans = null;
```
msweaver: can we keep this package scoped or protected?

PR #600, opened 2013-10-23
CONN-645:Fix SAML Issuer when Format is Entity and SPProvidedId is given

alameluchidambaram:


Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/saml/extraction/SamlTokenCreator.java
```diff
@@ -159,6 +159,16 @@
...
             }
+            
+            if(assertion.getSamlIssuer()!= null) {
+                if(NullChecker.isNotNullish(assertion.getSamlIssuer().getIssuer())) {
+                    requestContext.put(NhincConstants.ASSERTION_ISSUER_PROP, assertion.getSamlIssuer().getIssuer());
+                }
+                if(NullChecker.isNotNullish(assertion.getSamlIssuer().getIssuerFormat())) {
+                     requestContext.put(NhincConstants.ASSERTION_ISSUER_FORMAT_PROP, assertion.getSamlIssuer()
+                        .getIssuerFormat());
+            }
```
msweaver: this is not properly formatted.

alameluchidambaram: Thx Matt! I will fix it.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/callback/cxf/CONNECTSamlAssertionValidator.java
```diff
@@ -118,13 +120,36 @@ protected void validateAssertion(AssertionWrapper assertion) throws WSSecurityEx
...
+                }
+
+                if (!StringUtils.isBlank(issuer.getSPNameQualifier())) {
+                    attributeNotRequired = issuer.getSPNameQualifier();
+
+                }
+            }
+            if (NullChecker.isNotNullish(attributeNotRequired)) {
+
+                throw new WSSecurityException("SOAP header element Security/Assertion/Issuer/@Format = " + ""
```
msweaver: I think this error message is a little misleading. Can we just throw the exception 3 times from each of the above if conditions? That way you can specifically say SPProviderID (or NameQualifier, or SPNameQualifier) cannot be present if issuer format is entity.

alameluchidambaram: I can throw for each one of the field specifically.

PR #611, opened 2013-11-07
CONN-694:  Fix Name in UC Correlation

jsmith2:
Updated so name is from business entity name and not the HCID from the key value.


jsmith2:
I found an issue with HomeCommunityMap test property accessor.  It looks like it might have a dependency on a prior run unit tests to have set up the property accessor.  I changed it so that it can be mocked correctly within that class for testing.

PR #613, opened 2013-11-12
CONN-674: Fixed Privacy Viloation issues

nareshcgi:


jsmith2:
Are these all the fixes for the privacy part?  The ticket calls out 26 issues and I count 13 fixes here.

nareshcgi:
So there are 14 issues(some of the issues are duplicate) and out of that 13 are fixed and the one related to CXFPasswordCallbackHandler needs more analysis, which is out of scope for this ticket.

PR #635, opened 2013-12-16
CONN-200: Added logic to send Failed MDN

nareshcgi:


Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/DirectReceiverImpl.java
```diff
@@ -109,19 +120,29 @@ public void receiveInbound(MimeMessage message) {
...
         }
-
+        //send the Processed MDN notification
         sendMdnProcessed(result);
 
         DirectEdgeProxy proxy = getDirectEdgeProxy();
-        proxy.provideAndRegisterDocumentSetB(processedEnvelope.getMessage());
+        boolean notificationToEdgeFailed = false;
+
+        try {
```
msweaver, line 129: Is this pull request still valid?

I have some questions about this logic. Isn't there a setting to see if CONNECT is configured to send mdns to the edge or not? I was expecting to see that logic here. Im also wondering if we could simply this a little.

PR #660, opened 2014-01-30
version 10.0.0 was not found at the forgerock repo, updated to 10.0.1

bhumphrey:
I didn't have org.forgerock.openam.clientsdk version 10.0.0 in my local repo. This version (10.0.0) appears to nolonger be available at the repo location: http://maven.forgerock.org/repo/releases.

This pull also includes cherry-picks (58bd8ed and 51070d9)  fixing the certs for testing.


msweaver:
Thanks Brian!

We did already make this change in the CONNECT/CONNECT_integration code line (but using version 10.0.1). I want to keep the versions consistent between the two branches. We are fixing to do a 4.2.1 release and we'll be sure to include this fix as part of that.

msweaver:
add to whitelist

PR #681, opened 2014-02-12
Conn 820

jsmith2:
Added junit tests for DirectCore xdr package unit tests.
- New unit test for DirectHeaderExtractor
- Updated SoapDirectEdgeOrchestrationTest to actually test the public method and corrected it's mocking and injection.


Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/xdr/DirectHeaderExtractor.java
```diff
@@ -91,9 +91,6 @@ protected SoapEdgeContext getHeaderProperties(WrappedMessageContext wMsgContext)
         }
         Message msg = wMsgContext.getWrappedMessage();
 
-        if (msg == null) {
```
msweaver, line 94: why take this out?

jsmith2, line 94: It's impossible for message to be null.  it would throw a NPE in the WrappedMessageContext if it were (it gets passed into the constructor) and therefore WrappedMessageContext would be null.

PR #683, opened 2014-02-13
CONN-821:Add unit Tests to Mail package of DirectCore

alameluchidambaram:


Product/Production/Services/DirectCore/src/test/java/gov/hhs/fha/nhinc/mail/ImapMailReceiverTest.java
```diff
@@ -0,0 +1,117 @@
...
+        when(mockFolder.getMessageCount()).thenReturn(1);
+        when(mockFolder.getMessages(1, 1)).thenReturn(msgs);
+        ImapMailReceiver receiver = new ImapMailReceiver(properties) {
+            @Override
+            protected Folder getInbox(Store store) {
+               return mockFolder; 
+            }
+        };
+        result = receiver.handleMessages(handler);
+        assertEquals(result,0);
```
jsmith2: Can we verify that the delete flag was called?

Maybe do: verify(message).setFlag(Flags.Flag.DELETED, true);

PR #691, opened 2014-02-19
CONN-853 Added last couple of unit tests for final coverage

jsmith2:
Updated for coverage.  Now over 80% unit coverage on DirectCore.

### DIFF
--- a/Product/Production/Services/DirectCore/src/test/java/gov/hhs/fha/nhinc/direct/edge/proxy/DirectEdgeProxySoapImplTest.java
+++ b/Product/Production/Services/DirectCore/src/test/java/gov/hhs/fha/nhinc/direct/edge/proxy/DirectEdgeProxySoapImplTest.java
@@ -26,8 +26,9 @@
  */
 package gov.hhs.fha.nhinc.direct.edge.proxy;
 
+import gov.hhs.fha.nhinc.direct.DirectBaseTest;
+import gov.hhs.fha.nhinc.direct.DirectException;
 import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
 import static org.mockito.Matchers.any;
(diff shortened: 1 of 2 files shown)


Product/Production/Services/DirectCore/src/test/java/gov/hhs/fha/nhinc/direct/edge/proxy/DirectEdgeProxySoapImplTest.java
```diff
@@ -69,61 +69,53 @@
...
-    @Test
-    @Ignore
-    public void testSend() throws MessagingException {
+    @Test(expected = DirectException.class)
+    public void testSendWithNoURL() throws MessagingException {
         MimeMessage message = new MimeMessage(null, IOUtils.toInputStream(DirectUnitTestUtil
                 .getFileAsString("Example_A.txt")));
         RegistryResponseType resp = objectUnderTest.provideAndRegisterDocumentSetB(message);
-
-        assertTrue(resp.getStatus().contains("Success"));
```
msweaver, line 109: Why remove this assertion?

jsmith2, line 109: It never gets hit.  objectUnderTest.provideAndRegister... throws a DirectException. See after @Test that it's expected.  That assertion would fail if it was hit, since the response would be null (actually it would throw a NPE).

msweaver, line 109: whoops, I didn't realize the exception was expected.

PR #697, opened 2014-03-07
CONN-903:Commit CXF TimestampCache propery disable feature in CONNECT

alameluchidambaram:


msweaver:
This change is only required for secured services. Unsecured services do not have timestamps.

alameluchidambaram:
ok.. will remove it from unsecured endpoints.

PR #700, opened 2014-03-11
CONN-892, CONN-908: Incorrect end point configured in UDDI and fixed POM...

vsaibabu:
... to point to 4.0.5 version of services...


msweaver:
Can you also update Product/SoapUI_Test/pom.xml from 4.0.4 to 4.0.5?

msweaver:
btw this is going to fail the PRV until maven central is updated.

PR #734, opened 2014-04-14
CONN-1045 Update PD Events for initiating/responding hcid

jsmith2:
Do not merge.  Will comment when ready.  Still testing with 2 gateways.


jsmith2:
Tested with 2 gateways, and both receiving and initiating gateways appeared to have the correct HCIDs in event logs.

Ready for merge.

PR #742, opened 2014-04-17
Conn 1020

alameluchidambaram:


Product/Production/Gateway/DocumentQuery_20/pom.xml
```diff
@@ -15,8 +15,35 @@
             <artifactId>DocumentQueryCore</artifactId>
             <version>${project.parent.version}</version>
         </dependency>
-        
-       
+		<dependency>
```
msweaver, line 18: why does this need to be added to DQ?

alameluchidambaram, line 18: when DQ starts the standard/Passthru and Outbound/Inbound classes also get instantiated which has dependency on hibernate jar since the Standard Outbound makes a database call.

PR #751, opened 2014-04-23
CONN-1072

christophermay07:
- Cleaned up references to HIH
- Scripts for creating and priming the Admin GUI database
- domain.xml additions for declaring datasource and connect pool
- POM updates for consistent deployment between app servers, and easier integration with CONNECT EAR in the future.


Product/Production/Adapters/General/CONNECTAdminGUI/src/main/webapp/WEB-INF/glassfish-web.xml
```diff
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
```
msweaver, line 1: isn't this file usually called sun-web.xml? What's the difference between sun-web and glassfish-web?

PR #761, opened 2014-04-28
CONN-1082

christophermay07:
- Fixed issue where Hibernate was not persisting new entities when id was preset.
- Removed flush statements in favor of letting container handle it automatically.
- Whitespace fixes (tabs to spaces, trimmed trailing space).

### DIFF
--- a/Product/Production/Adapters/General/CONNECTDirectConfig/src/main/java/gov/hhs/fha/nhinc/directconfig/dao/impl/AnchorDaoImpl.java
+++ b/Product/Production/Adapters/General/CONNECTDirectConfig/src/main/java/gov/hhs/fha/nhinc/directconfig/dao/impl/AnchorDaoImpl.java
@@ -31,7 +31,6 @@ import java.util.List;
 import java.util.Locale;
 
 import org.hibernate.Query;
-
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.hibernate.SessionFactory;
@@ -42,7 +41,7 @@ import gov.hhs.fha.nhinc.directconfig.exception.CertificateException;
 import gov.hhs.fha.nhinc.directconfig.dao.AnchorDao;
 
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.stereotype.Service;
+import org.springframework.stereotype.Repository;
 import org.springframework.transaction.annotation.Transactional;
 
 /**
@@ -50,7 +49,7 @@ import org.springframework.transaction.annotation.Transactional;
  *
  * @author ppyette
  */
-@Service
+@Repository
 public class AnchorDaoImpl implements AnchorDao {
 
     @Autowired
@@ -151,7 +150,7 @@ public class AnchorDaoImpl implements AnchorDao {
         log.debug("Enter");
 
         if (anchor != null) {
-        	anchor.setId(null);
+            anchor.setId(null);
             anchor.setCreateTime(Calendar.getInstance());
 
             try {
--- a/Product/Production/Adapters/General/CONNECTDirectConfig/src/main/java/gov/hhs/fha/nhinc/directconfig/dao/impl/AnchorDaoImpl.java
+++ b/Product/Production/Adapters/General/CONNECTDirectConfig/src/main/java/gov/hhs/fha/nhinc/directconfig/dao/impl/AnchorDaoImpl.java
@@ -1,52 +1,77 @@
 /*
- Copyright (c) 2010, NHIN Direct Project
- All rights reserved.
+ * Copyright (c) 2009-2014, United States Government, as represented by the Secretary of Health and Human Services.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *     * Redistributions of source code must retain the above
+ *       copyright notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the documentation
+ *       and/or other materials provided with the distribution.
+ *     * Neither the name of the United States Government nor the
+ *       names of its contributors may be used to endorse or promote products
+ *       derived from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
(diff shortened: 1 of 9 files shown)


Product/Production/Adapters/General/CONNECTDirectConfig/src/main/java/gov/hhs/fha/nhinc/directconfig/dao/impl/CertPolicyDaoImpl.java
```diff
@@ -316,19 +312,19 @@ public CertPolicyGroup getPolicyGroupById(long id) throws ConfigurationStoreExce
...
+	public void addPolicyGroup(CertPolicyGroup group) throws ConfigurationStoreException
 	{
 //    	try
 //    	{
 //			final CertPolicyGroup existingPolicyGroup = this.getPolicyGroupByName(group.getPolicyGroupName());
 //			if (existingPolicyGroup != null)
 //				throw new ConfigurationStoreException("Certificate policy group " + group.getPolicyGroupName() + " already exists");
-//			
+//
+//			group.setId(null);
```
msweaver: is this commented out? Why not just delete?

Product/Production/Adapters/General/CONNECTDirectConfig/src/main/java/gov/hhs/fha/nhinc/directconfig/dao/impl/DNSDaoImpl.java
```diff
@@ -43,106 +43,103 @@ Neither the name of The Direct Project (directproject.org) nor the names of its
  * @author Greg Meyer
  * @since 1.1
  */
-//@Repository
```
msweaver, line 46: Using this as a guide: http://stackoverflow.com/questions/1440096/difference-between-repository-and-service

is it still correct to remove @Repository? Do we do any manipulation of the objects or perform any service aside from storing/retrieving the data?

Product/Production/Adapters/General/CONNECTDirectConfig/src/main/java/gov/hhs/fha/nhinc/directconfig/dao/impl/AnchorDaoImpl.java
```diff
@@ -178,7 +179,6 @@ public void add(Anchor anchor) {
             log.debug("Calling JPA to persist the Anchor");
 
             sessionFactory.getCurrentSession().persist(anchor);
-            sessionFactory.getCurrentSession().flush();
```
msweaver, line 181: Why do we have to remove this?

Product/Production/Adapters/General/CONNECTDirectConfig/src/main/java/gov/hhs/fha/nhinc/directconfig/dao/impl/CertPolicyDaoImpl.java
christophermay07: Cert policy functionality is currently no-op'ed in our code, but the assumption was that it will eventually be implemented as part of CONNECT; by implementing the CONN-1082 "<entity>.setId(null)" fix here as well (even if just commented out), we don't have to track down the issue again when this class is actually implemented.

Product/Production/Adapters/General/CONNECTDirectConfig/src/main/java/gov/hhs/fha/nhinc/directconfig/dao/impl/AnchorDaoImpl.java
christophermay07, line 181: The flush statement is, at best, redundant; a flush happens at commit.  In the code we absorbed, persist was being called on the session without an explicit commit following, hence the flush.  With our switch to JTA-CMT, we let the container handle when to commit and flush; our code's only responsibility is to get the session properly.

Hibernate's documentation offers the following: "With CMT, transaction demarcation is completed in session bean deployment descriptors, not programmatically."
- http://docs.jboss.org/hibernate/core/3.3/reference/en/html/transactions.html#transactions-demarcation-jta

msweaver, line 181: ty

Product/Production/Adapters/General/CONNECTDirectConfig/src/main/java/gov/hhs/fha/nhinc/directconfig/dao/impl/CertPolicyDaoImpl.java
msweaver: IMO I would remove this code. We don't have plans to implement right now and our solution will likely need to work for exchange and direct so it's unlikely to follow this same implementation. We can always get the latest code from the RI again.

Product/Production/Adapters/General/CONNECTDirectConfig/src/main/java/gov/hhs/fha/nhinc/directconfig/dao/impl/DNSDaoImpl.java
christophermay07, line 46: Good catch; it appears both annotations will work in this case, but the Repository annotation is the one intended for use with a DAO.  I'll update the PR with the annotations changes for all DAO Impl classes.

Product/Production/Adapters/General/CONNECTDirectConfig/src/main/java/gov/hhs/fha/nhinc/directconfig/dao/impl/CertPolicyDaoImpl.java
christophermay07: Even deleting the commented lines, I assume we need to leave method stubs so that we'll be compliant with Direct's ConfigurationService wsdl.  Is this an acceptable change?

msweaver:
Sorry Chris, two more comments - this will be the best PR ever ;-)
1) Add the CONNECT license
2) Add/update java docs.

christophermay07:
@msweaver Per your comments above:

1) Added the CONNECT license.
2) Every class touched by this PR inherits its interface's javadocs (although I added all missing "Override" annotations to help make this clearer), whether a "non-Javadoc" comment has been added or the "inheritDoc" annotation has been declared explicitly.  For reference:  http://docs.oracle.com/javase/6/docs/technotes/tools/solaris/javadoc.html#inheritingcomments

PR #763, opened 2014-04-29
Change default start page to Login page in web.xml

kerry-chin:


msweaver:
CR complete. Waiting on PRV.

PR #764, opened 2014-04-30
Conn 1074

alameluchidambaram:


Product/Production/Adapters/General/gateway.properties
```diff
@@ -0,0 +1,128 @@
+# Duration in seconds between file reload
```
msweaver: why is this here?

msweaver:
There are a number of formatting issues with the poms specifically.

Product/Production/Adapters/General/gateway.properties
alameluchidambaram: Gateway.properties is added accidently here while dragging in windows explorer.will remove it from codebase.

alameluchidambaram:
Fixed all the formatting pom.xml issues using Eclipse.

PR #768, opened 2014-05-02
CONN-1087

christophermay07:
Set createTime to non-nullable, non-updatable for all entities

### DIFF
--- a/Product/Production/Adapters/General/CONNECTDirectConfig/src/main/java/gov/hhs/fha/nhinc/directconfig/entity/CertPolicy.java
+++ b/Product/Production/Adapters/General/CONNECTDirectConfig/src/main/java/gov/hhs/fha/nhinc/directconfig/entity/CertPolicy.java
@@ -18,19 +18,17 @@ import org.nhindirect.policy.PolicyLexicon;
 
 @Entity
 @Table(name = "certpolicy")
-public class CertPolicy 
-{
+public class CertPolicy {
     private Long id;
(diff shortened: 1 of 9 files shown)


Product/Production/Adapters/General/CONNECTDirectConfig/src/main/java/gov/hhs/fha/nhinc/directconfig/entity/CertPolicy.java
```diff
@@ -39,88 +37,76 @@ public CertPolicy()
...
     }
-    
+
     /**
      * Get the value of createTime.
      * 
      * @return the value of createTime.
      */
-    @Column(name = "createTime", nullable = false)    
+    @Column(updatable = false, nullable = false)
```
msweaver, line 98: noticed the name is gone, is this intentional and ok?

christophermay07, line 98: This was intentional; since Column name defaults to the property/field name, explicitly including it when they're equal is redundant at best: http://docs.oracle.com/javaee/6/api/javax/persistence/Column.html#name%28%29

Also, case-sensitivity is not an issue in this case, as column names are case insensitive on all platforms:  http://dev.mysql.com/doc/refman/5.1/en/identifier-case-sensitivity.html

PR #769, opened 2014-05-02
CONN-1009: DirectSender/DirectReceiver Proxy classes.

vsaibabu:


Product/Production/Services/DirectCore/nbactions.xml
```diff
@@ -0,0 +1,49 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
msweaver: this file, along with the catalog.xml should be removed from the change set and added to gitignore.

Product/Production/Services/DirectCore/src/jax-ws-catalog.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
msweaver: this is another netbeans file - remove and gitignore.

msweaver:
Two overall comments:
1) I think WSDLs/schemas should go in src/main/resources/wsdl and src/main/resources/schema respectively.
2) Typically web service ports are wrapped up in these client classes (https://github.com/CONNECT-Solution/CONNECT/blob/CONNECT_integration/Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/messaging/client/CONNECTCXFClientUnsecured.java) so that they are decorated with all the CONNECT messaging stuff. Im not sure if this is required without testing so... just an FYI.

vsaibabu:
Matt,
1 - WSDL/Schema's need to be cleaned once we have them in CONNECT-Webservices and CONNETCommonTypes project.
2 - I am unable to use CONNECTCXFClientUnsecured because these services doesn't have AssertionType element.
Created user story to support these comments and address the issues in coming sprints.https://connectopensource.atlassian.net/browse/CONN-1092

PR #771, opened 2014-05-05
Replace copyright HTML code in copyright file

kerry-chin:
XHTML:
- Changed copyright symbol code in copyright.xhtml file.

Notes:
- Using HTML Name "&copy;" was causing errors.
- Changed to HTML Number "&#169;" to fix.


msweaver:
This is good to be merged. Just waiting on the PRV.

PR #782, opened 2014-05-14
Creation time

dankmirth:
Context: http://connect-forums.3294226.n2.nabble.com/NwHIN-PD-can-have-invalid-creation-time-td7579571.html#a7579572

This is from the CONNECT 4.2 branch


dankmirth:
I'm going to create a new one starting from CONNECT_integration.  I'm not too familiar with github

PR #785, opened 2014-05-15
CONN:1090 Jquery Fix

sailajaa:


Product/Production/Adapters/General/CONNECTAdminGUI/src/main/webapp/resources/Templates/BaseTemplate.xhtml
```diff
@@ -36,11 +36,9 @@
     <h:head>
         <title><ui:insert name="title">Connect Administration</ui:insert></title>
         <ui:insert name="metatag">
-        </ui:insert>
-        <h:outputStylesheet library="css" name="bootstrap.css" />
```
jsmith2, line 40: why was this moved?  isn't it included the same either way?

PR #794, opened 2014-05-22
CONN-1119 & CONN-1143: Display total Inbound and Outbound transactions f...

vsaibabu:
...or Direct Messages


whitecat:
I think this commit message is for CONN-1134 not CONN-1143

PR #798, opened 2014-05-27
Fix Two More Issues from Admin GUI Improvements Wiki

kerry-chin:
This PR addresses two more notes found in the "Admin GUI Improvements" Wiki, found here:
https://connectopensource.atlassian.net/wiki/display/FCPT/Admin+GUI+Improvements

<b>SPECIFICALLY:</b>
- The Activity Logs screens should be converted to use the JSF templating structure.  <b>logs.xhtml code modified.</b>
- Modal box function: To cancel/close the modal pop-up box, the End User may click outside the modal box; however, this will cancel any changes they have made to the form contained inside the modal. To prevent loss of data if an End User accidentally clicks outside the modal, the modal programming needs to be changed so the click event handler will ignore clicks outside of the modal.  <b>For all instances where the modal box contained a form, clicking outside the box is now ignored.  This impacts the direct.xhtml and direct_domains_create.xhtml files.</b>


Product/Production/Adapters/General/CONNECTAdminGUI/src/main/webapp/resources/css/pf-override.css
```diff
@@ -0,0 +1,232 @@
+@charset "utf-8";
+/* CSS Document */
+
+/* == TABLE OF CONTENTS ===================================================== */
+/* == TABLE OF CONTENTS ===================================================== */
+/* == TABLE OF CONTENTS =====================================================
```
christophermay07, line 6: Quick question -- why don't lines 4 & 5 TOC have any content, but the one starting with line 6 does?

kerry-chin, line 6: It should end further down after any listed items and notes.

:: Kerry

On Jun 2, 2014, at 6:10 PM, "Christopher May" <notifications@github.com<mailto:notifications@github.com>> wrote:

In Product/Production/Adapters/General/CONNECTAdminGUI/src/main/webapp/resources/css/pf-override.css:

> @@ -0,0 +1,232 @@
> +@charset "utf-8";
> +/\* CSS Document _/
> +
> +/_ == TABLE OF CONTENTS ===================================================== _/
> +/_ == TABLE OF CONTENTS ===================================================== _/
> +/_ == TABLE OF CONTENTS =====================================================

Quick question -- why doesn't line 6 end with "*/" like the lines before it?

jsmith2:
Duplicated by PR-806

PR #801, opened 2014-05-30
Conn 1131 - Role Model and Services

jsmith2:
Allows for creation of user with role, update role values, and uses role for decision on page access.


Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/hibernate/dao/UserLoginDAOImpl.java
```diff
@@ -87,6 +93,95 @@ public boolean createUser(UserLogin createUser) {
...
+    @Override
+    public UserRole getRole(long role) {
+        Session session = null;
+        Transaction tx = null;
+        
+        UserRole result = null;
+        
+        try {
+            session = this.sessionFactory.openSession();
+            tx = session.beginTransaction();
```
christophermay07: Transaction is begun but not ended.  Is this intentional, should there be a commit with a rollback in the catch, or do we not need the transaction at all for read-only access?

Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/jee/jsf/UserAuthorizationListener.java
```diff
@@ -85,17 +95,19 @@ public void afterPhase(PhaseEvent event) {
...
         if (!noLoginRequiredPages.contains(currentPage) && currentUser == null) {
             LOG.debug("login required and current user is null, redirecting to login page.");
             NavigationHandler nh = facesContext.getApplication().getNavigationHandler();
             nh.handleNavigation(facesContext, null, LOGIN_PAGE_NAV_OUTCOME);
-        }else {
-            //TODO check role here.
+        }else if(currentUser != null && !roleService.checkRole(formatPageName(currentPage), currentUser)) {
+            LOG.debug("Current User does not have permission for page, redirecting to status page.");
+            NavigationHandler nh = facesContext.getApplication().getNavigationHandler();
+            nh.handleNavigation(facesContext, null, STATUS_PAGE_NAV_OUTCOME);
```
christophermay07, line 110: Does every user have access to the status page?  If not, what happens here when they try to access the status page?

jsmith2, line 110: I don't think we've decided on this behavior.  I think status page is just a placeholder for now.  I think we'll need to introduce a page error handling concept, my suggestion would be that we redirect back to the referring page and pop up a dialog box stating the denial  of access.  Need a ticket for this.

On Jun 1, 2014, at 7:46 PM, Christopher May wrote:

In Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/jee/jsf/UserAuthorizationListener.java:

> ```
>      }
> 
>      if (!noLoginRequiredPages.contains(currentPage) && currentUser == null) {
>          LOG.debug("login required and current user is null, redirecting to login page.");
>          NavigationHandler nh = facesContext.getApplication().getNavigationHandler();
>          nh.handleNavigation(facesContext, null, LOGIN_PAGE_NAV_OUTCOME);
> ```
> -        }else {
> -            //TODO check role here.
> -        }else if(currentUser != null && !roleService.checkRole(formatPageName(currentPage), currentUser)) {
> -            LOG.debug("Current User does not have permission for page, redirecting to status page.");
> -            NavigationHandler nh = facesContext.getApplication().getNavigationHandler();
> -            nh.handleNavigation(facesContext, null, STATUS_PAGE_NAV_OUTCOME);

Does every user have access to the status page? If not, what happens here when they try to access the status page?

PR #815, opened 2014-06-11
CONN-1161 (DirectHBM)

christophermay07:
Rewrote Direct config DAOs to use a Hibernate Mapping file instead of JPA annotations.


Product/Production/Adapters/General/CONNECTDirectConfig/src/main/java/gov/hhs/fha/nhinc/directconfig/dao/impl/AnchorDaoImpl.java
```diff
@@ -199,199 +192,213 @@ public void add(Anchor anchor) {
...
      */
     @Override
-    @Transactional(readOnly = false)
     public void save(Anchor anchor) {
-        if (anchor != null) {
-            List<Anchor> anchors = new ArrayList<Anchor>();
-            anchors.add(anchor);
-            save(anchors);
-        }
+        // Deleted due to non-terminating circular reference
```
jsmith2, line 222: Should we throw an unsupportedoperationexception for these deleted methods.  Or also annotate them as deprecated?

christophermay07, line 222: I'm hesitant to deprecate -- if the issue is fixed in a future version of Direct RI and we decide to absorb that code, we will have to "un-deprecate" the function.

As for UnsupportedOperationException, I've been using the following at the service level on unimplemented web services: "throw new ConfigurationServiceException(new NotImplementedException());"  -- I'll add the same for the two Anchor save functions, especially since they are not referenced anywhere at the moment.

Product/Production/Adapters/General/CONNECTDirectConfig/src/main/java/gov/hhs/fha/nhinc/directconfig/dao/impl/TrustBundleDaoImpl.java
```diff
@@ -257,36 +304,59 @@ public void updateLastUpdateError(long trustBundleId, Calendar attemptTime, Bund
...
 
-                this.disassociateTrustBundleFromDomains(id);
+                    for (long id : trustBundleIds) {
+                        try {
+                            final TrustBundle bundle = getTrustBundleById(id);
 
-                sessionFactory.getCurrentSession().delete(bundle);
-            } catch (ConfigurationStoreException e) {
-                log.warn(e.getMessage(), e);
+                            // disassociateTrustBundleFromDomains(id);
```
jsmith2: should this line just be deleted not commented out?

Product/Production/Adapters/General/CONNECTDirectConfig/src/main/java/gov/hhs/fha/nhinc/directconfig/dao/impl/TrustBundleDaoImpl.java
```diff
@@ -496,30 +636,98 @@ public void disassociateTrustBundleFromDomains(long trustBundleId) throws Config
...
+            session = DaoUtils.getSession();
+
+            if (session != null) {
+                log.debug("Saving " + anchors.size() + " Trust Bundle Anchors");
 
-        return retVal;
+                tx = session.beginTransaction();
+
+                for (TrustBundleAnchor tba : anchors) {
+                    // TODO: set trust bundle?
```
jsmith2: is this a chris todo or did this already exist as a todo in the direct code?

Product/Production/Adapters/General/CONNECTDirectConfig/src/main/java/gov/hhs/fha/nhinc/directconfig/dao/impl/TrustBundleDaoImpl.java
christophermay07: Yes, good catch!

Product/Production/Adapters/General/CONNECTDirectConfig/src/main/java/gov/hhs/fha/nhinc/directconfig/dao/impl/TrustBundleDaoImpl.java
christophermay07: Whoops.... this was a Chris todo that I forgot to remove after it became a Chris done!  Removed two other Chris todo's as well :)

PR #838, opened 2014-06-27
Conn 1164

jsmith2:
With Common-type and CONNECT-Webservice pull requests (merge last).

Need to change pom for correct common-types/webservice values after pushing to maven repo.


Product/Production/Adapters/General/CONNECTAdminGUI/pom.xml
```diff
@@ -17,6 +17,16 @@
             <version>${project.parent.version}</version>
         </dependency>
         <dependency>
+            <groupId>org.connectopensource</groupId>
```
jsmith2, line 20: Build will fail until other related commits are merged and uploaded maven repo (will also need to update version here).

whitecat:
Should this be linked to issue [CONN-1200](https://connectopensource.atlassian.net/browse/CONN-1200)? Meaning that the title shouldn't be 1164 but rather it should be 1200?

PR #856, opened 2014-07-19
CONN-1202

christophermay07:
Added TrustBundle-Domain Relation functionality


Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/direct/DirectTrustBundleBean.java
```diff
@@ -58,17 +60,17 @@
     private TrustBundle selectedTb;
 
     public List<TrustBundle> getTrustBundles() {
-        GetTrustBundles gtb = new GetTrustBundles();
-        gtb.setFetchAnchors(true);
-        return directService.getTrustBundles(gtb);
+        return directService.getTrustBundles(true);
```
jsmith2, line 63: Why don't you lazy load this like the directdomainbean with the refreshes?  won't this have the same multiple call issue.

christophermay07, line 63: CONN-1202 only covers the Domain backing bean, so I only added lazy-load logic there.  I think modifying any other backing beans should be part of CONN-1220 -- I've already included a comment on that ticket which points back to the relevant code from this PR.

PR #874, opened 2014-08-08
CONN-1280-Outbound

christophermay07:
Added Passthrough Outbound / Entity classes for CORE X12 DS Service Core


Product/Production/Services/CORE_X12DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/corex12/docsubmission/realtime/entity/OutboundCORE_X12DSRealTimeFactory.java
```diff
@@ -0,0 +1,50 @@
...
+import gov.hhs.fha.nhinc.orchestration.OrchestrationContextBuilder;
+
+/**
+ * @author cmay
+ *
+ */
+public class OutboundCORE_X12DSRealTimeFactory {
+
+    private static OutboundCORE_X12DSRealTimeFactory INSTANCE = new OutboundCORE_X12DSRealTimeFactory();
+
```
msweaver: Shouldn't we pattern singletons like this: http://stackoverflow.com/a/71683?

christophermay07: @msweaver I agree that we should pattern singletons differently, but is there a reason you prefer that over using an enum (the accepted answer from your link)?

PR #878, opened 2014-08-19
CONN-1299: Created Entity/Nhin GenericBatch Request/ Response services.....

vsaibabu:
CONN-1299: Added Entity GenericBatch request and response interfaces. Added NHIN Interface at Gateway level. After successful deployment of CORE X12 Gateway project now we should have access to all Generic Batch Webservices along with Realtime.


nareshcgi:
I am good with the changes.

PR #881, opened 2014-08-21
CONN-1298 Added Adapter Interfaces Secured/Unsecured..

vsaibabu:
Added Secured/Unsecured Generic Batch Request and Response interfaces for EDI X12 Adapter.


alameluchidambaram:
good with this pr.

PR #882, opened 2014-08-22
Conn 1312

sailajaa:


Product/Production/Services/CORE_X12DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/corex12/docsubmission/realtime/adapter/AdapterCORE_X12DSRealTimeOrchImpl.java
```diff
@@ -74,6 +74,8 @@ private COREEnvelopeRealTimeResponse buildAdapterCORE_X12DSRealTimeResponseMetad
         oResponse.setCORERuleVersion("2.2.0");
         oResponse.setPayload("PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0iVVRGLTgiPz4NCjxDbGluaWNhbERvY3VtZW50\n"
             + "IG1vb2RDb2RlPSJFVk4iIHhtbG5zPSJ1cm46aGw3LW9yZzp2MyINCiAgICB4bWxuczp4c2k9Imh0");
+        oResponse.setErrorCode("Success");
+        oResponse.setErrorMessage(null);
```
jsmith2: why do we need to explicitly set this to null.  shouldn't it be null by default?

PR #895, opened 2014-09-11
CONN:1328 Updated SOAP UI test case to include Relatesto and messageid in the gene...

sailajaa:
...ric batch submission deferred response


nareshcgi:
Reviewed looks good.

PR #898, opened 2014-09-17
Conn 1329

vsaibabu:
Implemented Large Payload functionality to send X12 Generic Batch request and response messages in more appropriate way with good performance. Since the payload is going to be large there needs to be followed Large Payload functionality and implemented it.


nareshcgi:
Looks good.

PR #902, opened 2014-09-20
CONN-1220: Updated Direct backing beans to have single calls to Direct C...

sailajaa:
...onfig services

### DIFF
--- a/Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/direct/DirectCertBean.java
+++ b/Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/direct/DirectCertBean.java
@@ -39,7 +39,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.stereotype.Component;
 
 /**
- * 
+ *
  * @author jasonasmith
  */
 @ManagedBean(name = "directCertBean")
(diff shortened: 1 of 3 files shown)


Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/direct/DirectCertBean.java
```diff
@@ -92,6 +84,7 @@ public void addCertificate() {
         certificate.setStatus(EntityStatus.NEW);
         addCert.getCerts().add(certificate);
         directService.addCertificate(addCert);
+        refreshCertificates();
```
jsmith2, line 87: Should this be here or where the action is called?  why are we refreshing when you upload a file to the UI?

jsmith2, line 87: Ignore comment.  I missed which method this was called from.  diff threw me off.

sailajaa, line 87: My Understanding is, I am refreshing to get list of certificates after add certificates action.

vsaibabu, line 87: I am good with this PR, Merging it now.

PR #915, opened 2014-10-07
CONN-1363

christophermay07:
Fixed NPE points of failure for AdminGUI Cert & Trust Bundle tabs

### DIFF
--- a/Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/direct/DirectCertBean.java
+++ b/Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/direct/DirectCertBean.java
@@ -25,8 +25,10 @@ import gov.hhs.fha.nhinc.admingui.model.direct.DirectCertificate;
 import gov.hhs.fha.nhinc.admingui.services.DirectService;
 import java.util.ArrayList;
 import java.util.List;
+import javax.faces.application.FacesMessage;
 import javax.faces.bean.ManagedBean;
 import javax.faces.bean.ViewScoped;
+import javax.faces.context.FacesContext;
 import org.nhind.config.AddCertificates;
 import org.nhind.config.Certificate;
 import org.nhind.config.CertificateGetOptions;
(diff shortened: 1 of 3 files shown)


Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/direct/DirectCertBean.java
```diff
@@ -80,7 +80,11 @@ public void certFileUpload(FileUploadEvent event) {
     public void addCertificate() {
         AddCertificates addCert = new AddCertificates();
         Certificate certificate = new Certificate();
-        certificate.setData(certFile.getContents());
+
+        if (certFile != null) {
```
jsmith2, line 83: Can you insert a cert with no data?  Seems like we should just not add a cert here and error back.  or make the cert a required field.

christophermay07, line 83: This isn't about a cert with no data, but rather a safety net in case the GUI is somehow triggering this function despite certFile being null.

jsmith2, line 83: Understood, but you still add the cert despite the certfile being null.  Does the config service even insert this entry or would it fail after the call?

christophermay07, line 83: Totally misunderstood your comment, makes total sense now.  Updating to make cert a required field in the front-end, but also adding a check in this function just in case.

Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/direct/DirectCertBean.java
```diff
@@ -78,13 +80,19 @@ public void certFileUpload(FileUploadEvent event) {
...
+            AddCertificates addCert = new AddCertificates();
+            Certificate certificate = new Certificate();
+            certificate.setData(certFile.getContents());
+            certificate.setStatus(EntityStatus.NEW);
+            addCert.getCerts().add(certificate);
+            directService.addCertificate(addCert);
+            refreshCertificates();
+        } else {
+            FacesContext.getCurrentInstance().addMessage("domainDeleteError",
```
jsmith2: so using the domainDeleteError message is fine, but really should be renamed now to reflect that it's used for multiple purposes.  maybe just "domainError".

nareshcgi:
retest this please

nareshcgi:
retest this please

PR #925, opened 2014-10-15
Conn 1384

alameluchidambaram:
CONN-1384: Exclude Direct War from CONNECT Ear packaging


Product/Production/Deploy/ear/jboss7/pom.xml
```diff
@@ -9,9 +9,60 @@
...
+                                    <artifactId>CONNECTDirectConfig</artifactId>
+                                    <contextRoot>/CONNECTDirectConfig</contextRoot>
+                                    <excluded>${direct.excluded}</excluded>
+                                </webModule>
+                                <webModule>
+                                    <groupId>org.connectopensource</groupId>
+                                    <artifactId>Direct</artifactId>
+                                    <contextRoot>/Direct</contextRoot>
+                                    <excluded>${direct.excluded}</excluded>
+                                    <excluded>true</excluded>
```
christophermay07, line 49: Should not have both "excluded" tags here

alameluchidambaram, line 49: I can remove the exclusion of direct dependency since Direct war is not packaged within ear.

nareshcgi:

PR #926, opened 2014-10-16
CONN-1197: Added logic to support new event logging events

nareshcgi:
1. Added logic to support new Direct event logging events
2. Updated the even logic code to add parent message id to the json description
3. Updated the event logic code to add parent message id to event transaction id field
4. Added parent message id to the QOS notifications
5. Added logic to handle inbound security processing errors


sailajaa:
I am good with PR to merge.

alameluchidambaram:
I am good with this PR

PR #930, opened 2014-10-21
CONN-970 - Log Forging

christophermay07:
Handled High and Low Fortify findings for "Log Forging" and "Log Forging (debug)"


Product/Production/Adapters/General/CONNECTDirectConfig/src/main/java/gov/hhs/fha/nhinc/directconfig/service/impl/DomainServiceImpl.java
```diff
@@ -98,7 +94,7 @@ public void init() {
     @Override
     public AddDomainResponse addDomain(AddDomain domain) throws ConfigurationServiceException {
         dao.add(domain.getDomain());
-        log.info("Added Domain: " + domain.getDomain().getDomainName());
+        log.info("Domain added successfully");
```
jsmith2, line 97: Not sure why every other direct config file here uses logging level debug for statements like this and this is level info.  Inconsistent (but out of scope).

jsmith2, line 97: Reviewed.  good with merge.

nareshcgi:

PR #935, opened 2014-10-22
CONN-1373 Fix for multiple entries for page for page preferences

jsmith2:
Issue was we were doing an unnecessary database query for preferences.  Preferences were already included in the role object.


christophermay07:

nareshcgi:

PR #937, opened 2014-10-23
CONN-1416:Fix Outbound Dispatch Notifications not to be send when sender...

alameluchidambaram:
Fix not to log failed notification for Outbound MDN Dispatch event in the case of invalid user of an appropriate domain when sender is not requesting notification.


nareshcgi:
Add the PR description.

alameluchidambaram:
Removing DIRECT_ERROR does not affect since that method is called by MessageMonitoringAPI and an event is already logged when Message Notification fails.

nareshcgi:

PR #941, opened 2014-10-29
CONN-1424:Create Oracle SQL scripts for configdb,adminguidb and messagem...

alameluchidambaram:
CONN-1424: Create Oracle SQL scripts for configdb,adminguidb and messagemonitoringdb tables.


nareshcgi:
Verified PR and looks good.

PR #946, opened 2014-10-29
CONN-1441

christophermay07:
- Changed password salt to Base64
- Added password confirmation field


jsmith2:
Reviewed and good with merge.

PR #950, opened 2014-10-31
Conn 1446: Verify working in WebSphere

jsmith2:
Reverted previous commit due to issues with spring proxy loading in config service impl.  Added synchronization for instantiating of service instead.

Tested in WAS, Weblogic, and Glassfish.


sailajaa:
I reviewed changes and good with PR merge.

christophermay07:

PR #953, opened 2014-11-04
CONN-1444: Modified the code to capture inbound Direct message Error in event logs

nareshcgi:
Modified the code to capture inbound Direct message Error in event logs.


Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/DirectReceiverImpl.java
```diff
@@ -128,16 +128,20 @@ public void receiveInbound(MimeMessage message) {
         MessageProcessResult result;
         //Catch all the errors (can be a security error or any other error)and drop the message for now
         //TODO: Add a retry logic before dropping the message in the future
+        boolean isMdn = MessageMonitoringUtil.isMdnOrDsn(message);
         try {
+            if (!isMdn){
```
jsmith2, line 133: shouldn't this be an if/else statement with the if(isMdn) statement below?

PR #954, opened 2014-11-05
CONN-1444: Reverted back part of the event logging code

nareshcgi:
Reversed the event logging code for BEGIN_INBOUND_DIRECT. Since we can't find if its a MDN or DIrect message before processing the message. On an Inbound Message failure, we will see only a DIRECT_ERROR event.


christophermay07:

PR #960, opened 2014-11-12
CONN-1375 Added ability to display all appserver info in admin gui

jsmith2:
I tested it locally and works with glassfish, WAS, and WLS.

Added 100 length limit on info since WAS was spitting out all patch information.


nareshcgi:
Good with the PR

christophermay07:

PR #973, opened 2014-12-12
CONN-1574: Removed Regression Suite tests Gateway-2085 and CONN-997

nareshcgi:
Removed below Regression Suite tests
1. Gateway-2085 (from Bimodal)
2. CONN-997 (from Manual)


alameluchidambaram:
Reviewed and looks good.

PR #974, opened 2014-12-12
CONN-1573: Removed CONN-740 from the Manual regression suite folder

nareshcgi:
Removed CONN-740 from the Manual regression suite folder.


alameluchidambaram:
Reviewed and looks good.

PR #981, opened 2014-12-19
FHAC-24: Removed ValidateSAMLResourceURIAttributeTest from the pom and also renamed the project name

nareshcgi:
Removed ValidateSAMLResourceURIAttributeTest from the pom
Renamed the project name from CONN-560 to ValidateSAMLResourceURIAttributeTest


jsmith2:
Looks good to me.  Do we need to wait for the validator?  Will this test be run with the validator or with the merge?

PR #983, opened 2014-12-22
FHAC-29

christophermay07:
Added mustUnderstand test cases for X12 (g0 passthrough), updated / clarified mustUnderstand manual step text.  Also added negative mustUnderstand test cases for DQ Standard, g0 and g1.


jsmith2:
Unable to code review test due to size.  Good with merge and will test once merged.

PR #984, opened 2014-12-22
FHAC-20

christophermay07:
- Added OWASP Dependency Check plugin to POM
- Removed DirectCore dependencies from non-Direct profile due to reporting failure with fresh .m2 repository


nareshcgi:

PR #993, opened 2015-01-12
FHAC-64:Revert Mail jar removal and package mail jar within CONNECT ear

alameluchidambaram:
Revert Mail jar removal and package mail jar within CONNECT ear

### DIFF
--- a/Product/Production/Deploy/ear/weblogic/pom.xml
+++ b/Product/Production/Deploy/ear/weblogic/pom.xml
@@ -19,11 +19,6 @@
         </dependency>
         <dependency>
             <groupId>org.apache.geronimo.specs</groupId>
-            <artifactId>geronimo-jaxws_2.2_spec</artifactId>
-            <version>1.1</version>
-        </dependency>
-        <dependency>
-            <groupId>org.apache.geronimo.specs</groupId>
             <artifactId>geronimo-ws-metadata_2.0_spec</artifactId>
             <version>1.1.3</version>
         </dependency>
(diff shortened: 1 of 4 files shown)


Product/Production/Deploy/ear/weblogic/pom.xml
```diff
@@ -19,11 +19,6 @@
         </dependency>
         <dependency>
             <groupId>org.apache.geronimo.specs</groupId>
-            <artifactId>geronimo-jaxws_2.2_spec</artifactId>
```
jsmith2, line 22: Does this need to be reverted?

alameluchidambaram, line 22: This jar needs to be removed as per OWASP dependency documentation.
https://cgiinterop.atlassian.net/wiki/display/FCPT/OWASP+Dependency+Check

jsmith2, line 22: Ok.  so you're removing this in addition to reverting the mail jars?  Just confusing since this wasn't  part of the description.

alameluchidambaram, line 22: In other Application servers this jar was added by cxf bundle. so the exclusion from cxf-bundle helped to remove this jar. But in Weblogic this jar is defined as a dependency in Weblogic specific pom.xml. Also it was tested by deploying the ear to Weblogic Application server.

nareshcgi:
Reviewed PR and looks good.

PR #996, opened 2015-01-15
FHAC-99: Changed the MinimumDegreeMatch value tag in ValidationSuite PD g0 and g1 requests

nareshcgi:
Changed the MinimumDegreeMatch value tag in ValidationSuite PD g0 and g1 requests.


jsmith2:
retest this please

PR #1010, opened 2015-01-30
FHAC-118

christophermay07:
- References newest Common-Types and CONNECT-Webservices versions in Product/pom.xml
- Updated CONNECTAdminGUI imports based on new package spaces from Common-Types and CONNECT-Webservices
- Added Direct war into JBoss 7 ear for Direct builds


nareshcgi:

PR #1016, opened 2015-02-18
Ping prop POC

jsmith2:
-- DO NOT MERGE -- Until Dependencies are merged and a full team code review is complete.
- Migration to Apache Configurations (adds file refresh formatting capabilities)
- Added display for AdminGUI with edit capabilities.
- Updated some unit tests that had poor property file coupling.  (had to leave some coupling for unit tests as it would require a major refactor)
- Code cleanup, including removing unnecessary file name caching.
- Bonus: Fixed issue with caching of user login value (see LoginBean change).


Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/PropertyBean.java
```diff
@@ -0,0 +1,139 @@
...
+        }
+    }
+
+    private void setProperties() {
+        properties = new ArrayList<PropValue>();
+        try {
+            Properties props = PropertyAccessor.getInstance().getProperties("gateway");
+            
+            if (props != null) {
+                for (Object key : props.keySet()) {
```
christophermay07, line 92: Why are our prop keys of type Object (instead of String)?

jsmith2, line 92: Properties could be of many types including int, boolean, double, etc.

Product/Production/Common/Properties/src/main/resources/gateway.properties
```diff
@@ -110,8 +109,9 @@ CoreX12GenericBatchTimeStampTimeToLive=300
 # This is only valid for  Core X12 Batch request and response services.
 CoreX12GenericBatchFutureTimeToLive=60
 
-#           Direct Message Monitoring properties
-#   ****These properties should be moved to DIRECT Config Service setting****
+!           Direct Message Monitoring properties
+!   ****These properties should be moved to DIRECT Config Service setting****
```
christophermay07, line 113: Why do these comments use "!" instead of "#"?

jsmith2, line 113: For display purposes.  I filter out "!" from displayed comments for each property value.  The admin gui uses the "#" values for the description of the properties.  Most of the "!" values are either commented out properties or group descriptions.

Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/PropertyBean.java
christophermay07, line 92: The **values** can be of any type, but wouldn't the **keys** always be of type String?  For example:

Key: UDDIRefreshActive (String)
Value: false (boolean)

jsmith2, line 92: the key set returns the type Set<Object>, I cast on each iteration.  can't cache on the set.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/transform/subdisc/HL7QueryParamsTransforms.java
```diff
@@ -53,12 +53,14 @@
  * @author Jon Hoppesch
  */
 public class HL7QueryParamsTransforms {
```
nareshcgi, line 55: Just a suggestion, HL7MessageIdGenerator.generateHL7MessageId() used in multiple transformer class, we can create a parent class with a wrapper method and use it in the sub class. Also change the HL7MessageIdGenerator to a singleton class. We can address this in future re-factoring effort.

nareshcgi:

PR #1019, opened 2015-02-24
FHAC-169:Fix Oracle scripts and set DatabaseEventLogger eventTime

alameluchidambaram:
Fix Oracle sql scripts in the following files
1. nhincdb_oracle.sql
   2.populateTestData_oracle.sql

Fix Mysql Sql scripts for event table in nhincdb.sql
set DataBaseEventLogger eventTime .


alameluchidambaram:
The id field default is specified not null and it has autoincrement on that field. so insertScripts for MySQL scripts does not require that prefid to be explicitly specified.

sailajaa:
Reviewed and good with PR.

PR #1020, opened 2015-02-24
FHAC-160 Updated Direct pom to include reporting during nightly build

sovanncgi:
Added properties to Direct pom.xml to include reports when automated Direct test is run nightly


sovanncgi:
Not needed

PR #1022, opened 2015-02-26
FHAC-174: CXF Unbundling

christophermay07:
- "skip.unpack" config prop for Maven dependency plugin added to fix MDEP-98 error when running "dependency:analyze" goal.
- java.util.logging references removed to prevent library conflicts after CXF-unbundling.
- Pre-unbundling cleanup ONLY used to remove unused dependencies / declare used dependencies.  Library versions were not changed otherwise.

### DIFF
--- a/Product/Production/Adapters/AdminDistribution_a0/pom.xml
+++ b/Product/Production/Adapters/AdminDistribution_a0/pom.xml
@@ -20,6 +20,18 @@
             <artifactId>AdminDistributionCore</artifactId>
             <version>${project.parent.version}</version>
         </dependency>
+        <dependency>
+            <groupId>org.connectopensource</groupId>
+            <artifactId>CONNECTCoreLib</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.connectopensource</groupId>
+            <artifactId>CONNECTCommonTypesLib</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.connectopensource</groupId>
+            <artifactId>AdminDistributionWebservices</artifactId>
+        </dependency>
     </dependencies>
 
     <build>
--- a/Product/Production/Adapters/General/CONNECTAdminGUI/pom.xml
+++ b/Product/Production/Adapters/General/CONNECTAdminGUI/pom.xml
@@ -14,19 +14,24 @@
     <modelVersion>4.0.0</modelVersion>
 
     <dependencies>
+        <!-- Java EE APIs -->
+        <dependency>
+            <groupId>javax</groupId>
+            <artifactId>javaee-api</artifactId>
+        </dependency>
+
         <!-- CONNECT Modules / Web Services -->
         <dependency>
             <groupId>org.connectopensource</groupId>
             <artifactId>CONNECTCoreLib</artifactId>
-            <version>${project.parent.version}</version>
-        </dependency>
-        <dependency>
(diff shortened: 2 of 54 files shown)


Product/Production/Adapters/General/CONNECTAdminGUI/pom.xml
```diff
@@ -86,43 +95,92 @@
...
             <groupId>javax.servlet</groupId>
             <artifactId>jstl</artifactId>
+            <scope>runtime</scope>
         </dependency>
         <dependency>
-            <groupId>javax.servlet</groupId>
-            <artifactId>servlet-api</artifactId>
+            <groupId>org.glassfish</groupId>
+            <artifactId>javax.servlet</artifactId>
         </dependency>
```
nareshcgi, line 167: Just curious, does the group id changed for this?

Product/Production/Adapters/AdminDistribution_a0/pom.xml
```diff
@@ -20,6 +20,18 @@
...
+        </dependency>
+        <dependency>
+            <groupId>org.connectopensource</groupId>
+            <artifactId>CONNECTCommonTypesLib</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.connectopensource</groupId>
+            <artifactId>AdminDistributionWebservices</artifactId>
+        </dependency>
     </dependencies>
```
nareshcgi, line 35: Trying to understand how it worked without these referenced libraries before (even in some of the others poms that changed)?

Product/Production/Adapters/General/CONNECTAdminGUI/pom.xml
christophermay07, line 167: We use servlet 2.5 in some places (javax.servlet:servlet-api), and servlet 3.0 in other places (org.glassfish:javax.servlet).  In both cases, they're provided, so this is more for consistency.

Product/Production/Adapters/AdminDistribution_a0/pom.xml
christophermay07, line 35: Originally, every module under "Adapters" inherited CONNECTCoreLib and CONNECTCommonTypesLib (amongst other dependencies) from the Adapters POM.  This is a bad practice (POM modules don't have dependencies), and was causing problems during CXF unbundling due to transitive dependencies.

Instead, each POM now declares, to the best of my ability, exactly what it needs -- no more, no less.

Product/Production/Adapters/General/CONNECTAdminGUI/pom.xml
```diff
@@ -86,43 +95,92 @@
         <!-- Spring framework -->
         <dependency>
             <groupId>org.springframework</groupId>
-            <artifactId>spring-core</artifactId>
```
nareshcgi, line 89: Any specific reason we are removing the Spring-core dependency?

christophermay07, line 89: Our code doesn't directly use spring-core packages (org.springframework.core.*).  That jar is a transitive dependency to spring-beans, however, so the jar is still included in the lib folder in the final build.

nareshcgi:

jsmith2:
Good with PR with no comments.

alameluchidambaram:
Reviewed PR and looks good.

PR #1023, opened 2015-02-26
FHAC-186 Null value for current user logged in

jsmith2:
Login bean was changed to view scope and the value that is looked up is null after login.  Created method to pull user name from session attribute.


christophermay07:
I've reviewed this PR, endorsing for merge without comment.

alameluchidambaram:
Reviewed PR and looks good.

TabassumJafri:
Reviewed PR .. looks good

PR #1038, opened 2015-03-31
FHAC-242: added Accept-Encoding:deflate to request header for pulling co...

TabassumJafri:
added Accept-Encoding:deflate to request header for pulling conformance from HAPI server


jsmith2:
Looks good.  Will merge after it goes through CI.

PR #1039, opened 2015-04-01
fixed the Read column in conformance table to get values from confResour...

TabassumJafri:
Fixed fhir.xhtml: the read column in conformance table now get vlaues from confResource.supportingRead instead of confResource.SupportingUpdate


jsmith2:
PR loooks good.  Will merge after validation.

PR #1047, opened 2015-04-08
FHAC-257:update AuditLogging-Passthrough Suite to verify all xml elements/attributes from Message field for X12Realtime testcase

sailajaa:
Update X12Realtime test case with verify message script for Nhin Inbound and Outbound messages
added auditlogging properties with xml attribute entries.


christophermay07:

PR #1051, opened 2015-04-22
FHAC-269:Update AuditLogging-Passthrough to verify xml attributes for X12BatchRequest

sailajaa:
Modified X12BatchRequest test case to add verify checks for xml attributes.
Modified X12GenericBatch test case to check activeparticipant elements based on roleidcode displayname.


nareshcgi:

PR #1055, opened 2015-04-29
FHAC-232 Added Inbound Direct tests to automated Direct test suite

sovanncgi:
Made changes to the following:
- Outbound Direct test case is now testing both outbound AND inbound. Assertions have been added for remote database to check inbound Direct events
- Test case for Direct testing without delivery notification also added to test suite

Will need to disable Direct testing on Jenkins until FHAC-263 is completed


nareshcgi:

PR #1058, opened 2015-05-07
FHAC-278: Fortify fix to XML External entity injection

sailajaa:
xml external entity injection fix : Updated TransformerFactory to do secure processing.


nareshcgi:

PR #1061, opened 2015-05-14
FHAC-300 Adapter Props

jsmith2:
Added adapter properties tab and some navigation for it.

Formatted adapter.properties for admingui view.


nareshcgi:
I am good with the PR.

PR #1064, opened 2015-05-18
FHAC-235:Automation of Direct TestCases for HappyPath testcases using anchors,trustbundle anchors and keystore anchors

alameluchidambaram:
1. The testcases were modified to get messageId from MessageMonitoring tables. 
2. Subject for each email is made unique by adding dateTime to Subject.
3. TrustBundle sql was fixed to remove trustbundle anchor appropriately so that keystore anchors can be picked.
4. Remove Setting table data for anchors/trustbundle so that Keystore anchor data can be picked up.


nareshcgi:
Good with the PR.

PR #1070, opened 2015-05-22
FHAC-228/FHAC-229: Added audit logging feature for X12 Batch Request and Response services

nareshcgi:
Fixed the Audit message transformation logic for Core X12 request and response services


Product/Production/Services/CORE_X12DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/corex12/docsubmission/audit/transform/CoreX12AuditDataTransform.java
```diff
@@ -0,0 +1,467 @@
...
+            AuditDataTransformConstants.EVENT_ID_CODE_SYS_NAME_DOC, CoreX12AuditDataTransformConstants.ACTIVE_PARTICPANT_ROLE_CODE_SOURCE_DISPLAY_NAME));
+        participant.setUserIsRequestor(isRequesting ? Boolean.TRUE : Boolean.FALSE);
+        return participant;
+    }
+
+    protected static AuditMessageType.ActiveParticipant getActiveParticipantDestination(NhinTargetSystemType target) {
+        String strUrl = null;
+        String strHost = null;
+        AuditMessageType.ActiveParticipant participant = new AuditMessageType.ActiveParticipant();
+        if (null != target) {
```
jsmith2, line 180: can we reverse these.  typically we use object != null.  that reads weird.

Product/Production/Services/CORE_X12DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/corex12/docsubmission/audit/transform/CoreX12AuditDataTransform.java
```diff
@@ -0,0 +1,467 @@
...
+ *
+ * @author nsubrama
+ */
+public class CoreX12AuditDataTransform {
+
+    private static final XDRTransforms oXDR = new XDRTransforms();
+
+    private static final Logger LOG = Logger.getLogger(CoreX12AuditDataTransform.class);
+
+    public static AuditSourceIdentificationType getAuditSourceIdentificationType() {
```
jsmith2, line 76: Making all of these methods static is going to make it hard to unit test and mock.

Product/Production/Services/CORE_X12DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/corex12/docsubmission/audit/transform/CoreX12AuditDataTransform.java
```diff
@@ -0,0 +1,467 @@
...
+     * @param userInfo
+     * @param userIsReq
+     * @return <code>AuditMessageType.ActiveParticipant</code>
+     */
+    private static AuditMessageType.ActiveParticipant createActiveParticipantFromUser(UserType userInfo,
+        Boolean userIsReq) {
+        String ipAddr = null;
+        AuditMessageType.ActiveParticipant participant = new AuditMessageType.ActiveParticipant();
+
+        if (ipAddr == null) {
```
jsmith2, line 106: unnecessary null check.  this value will always be null.  you set it right before here.

Product/Production/Services/CORE_X12DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/corex12/docsubmission/audit/transform/CoreX12AuditDataTransform.java
```diff
@@ -0,0 +1,467 @@
...
+        }
+
+        // Set the User Id
+        String userId = null;
+        if (userInfo != null && userInfo.getUserName() != null && userInfo.getUserName().length() > 0) {
+            userId = userInfo.getUserName();
+        }
+
+        if (userId != null) {
+            participant.setUserID(userId);
```
jsmith2, line 122: why are we doing this here, instead of inside the previous if statement.  unnecessary if statement.

Product/Production/Services/CORE_X12DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/corex12/docsubmission/audit/transform/CoreX12AuditDataTransform.java
```diff
@@ -0,0 +1,467 @@
...
+            if (userInfo.getPersonName().getFamilyName() != null
+                && userInfo.getPersonName().getFamilyName().length() > 0) {
+                if (userName != null) {
+                    userName += (" " + userInfo.getPersonName().getFamilyName());
+                } else {
+                    userName = userInfo.getPersonName().getFamilyName();
+                }
+            }
+        }
+        if (userName != null) {
```
jsmith2, line 141: same here.  could set this inside a previous if where we know username isn't null.

Product/Production/Services/CORE_X12DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/corex12/docsubmission/audit/transform/CoreX12AuditDataTransform.java
```diff
@@ -0,0 +1,467 @@
...
+import org.caqh.soap.wsdl.corerule2_2_0.COREEnvelopeBatchSubmissionResponse;
+import org.caqh.soap.wsdl.corerule2_2_0.COREEnvelopeRealTimeRequest;
+import org.caqh.soap.wsdl.corerule2_2_0.COREEnvelopeRealTimeResponse;
+
+/**
+ * Class provides generic Audit Data Transformation APIs utilized across all the services
+ *
+ * @author nsubrama
+ */
+public class CoreX12AuditDataTransform {
```
jsmith2, line 70: Can we clean up this class to have the method types in the right order.  typically you'd have all public methods at the top followed by protected and then private.

Product/Production/Services/CORE_X12DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/corex12/docsubmission/audit/transform/CoreX12AuditDataTransform.java
```diff
@@ -0,0 +1,467 @@
...
+            participant.setNetworkAccessPointID(ipAddr);
+            participant.setNetworkAccessPointTypeCode(AuditDataTransformConstants.NETWORK_ACCESSOR_PT_TYPE_CODE_IP);
+        } catch (UnknownHostException ex) {
+            LOG.warn("UnknownHostException thrown getting local host address.", ex);
+            return null;
+        }
+
+        participant.getRoleIDCode().add(AuditDataTransformHelper.createCodeValueType(CoreX12AuditDataTransformConstants.ACTIVE_PARTICPANT_ROLE_CODE_CDE, null,
+            AuditDataTransformConstants.EVENT_ID_CODE_SYS_NAME_DOC, CoreX12AuditDataTransformConstants.ACTIVE_PARTICPANT_ROLE_CODE_SOURCE_DISPLAY_NAME));
+        participant.setUserIsRequestor(isRequesting ? Boolean.TRUE : Boolean.FALSE);
```
jsmith2, line 172: unecessary if statement.  why are we checking the boolean instead of just setting: participant.setUserIsRequestor(isRequesting);

Product/Production/Services/CORE_X12DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/corex12/docsubmission/audit/transform/CoreX12AuditDataTransform.java
```diff
@@ -0,0 +1,467 @@
...
+            if (msg instanceof COREEnvelopeRealTimeRequest) {
+                oRequestNoPayload = (COREEnvelopeRealTimeRequest) msg;
+                oRequestNoPayload.setPayload("");
+                xmlqname = new javax.xml.namespace.QName(CoreX12AuditDataTransformConstants.CORE_X12_NAMESPACE_URI, CoreX12AuditDataTransformConstants.CORE_X12_REQUEST_LOCALPART);
+                JAXBElement<org.caqh.soap.wsdl.corerule2_2_0.COREEnvelopeRealTimeRequest> element = new JAXBElement<org.caqh.soap.wsdl.corerule2_2_0.COREEnvelopeRealTimeRequest>(xmlqname, COREEnvelopeRealTimeRequest.class, oRequestNoPayload);
+                marshaller.marshal(element, baOutStrm);
+                bObject = baOutStrm.toByteArray();
+            }
+
+            if (msg instanceof COREEnvelopeRealTimeResponse) {
```
jsmith2, line 283: these ifs should be else ifs.  currently this code will check every if statement.

Product/Production/Services/CORE_X12DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/corex12/docsubmission/audit/transform/CoreX12AuditDataTransform.java
```diff
@@ -0,0 +1,467 @@
...
+            if (msg instanceof COREEnvelopeBatchSubmissionResponse) {
+                oBatchResponseNoPayload = (COREEnvelopeBatchSubmissionResponse) msg;
+                oBatchResponseNoPayload.setPayloadID("");
+                xmlqname = new javax.xml.namespace.QName(CoreX12AuditDataTransformConstants.CORE_X12_NAMESPACE_URI, CoreX12AuditDataTransformConstants.CORE_X12_BATCH_RESPONSE_LOCALPART);
+                JAXBElement<org.caqh.soap.wsdl.corerule2_2_0.COREEnvelopeBatchSubmissionResponse> element = new JAXBElement<org.caqh.soap.wsdl.corerule2_2_0.COREEnvelopeBatchSubmissionResponse>(xmlqname, COREEnvelopeBatchSubmissionResponse.class, oBatchResponseNoPayload);
+                marshaller.marshal(element, baOutStrm);
+                bObject = baOutStrm.toByteArray();
+            }
+        } catch (JAXBException ex) {
+            throw new RuntimeException(ex.getCause());
```
jsmith2, line 310: do we want to throw a runtime exception for audit logging?  won't this kill the whole transaction.  auditing shouldn't cause an inbound/outbound message to fail.

Product/Production/Services/CORE_X12DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/corex12/docsubmission/audit/transform/CoreX12AuditDataTransform.java
```diff
@@ -0,0 +1,467 @@
...
+                participant.setUserID(strUrl);
+                strHost = url.getHost();
+                participant.setNetworkAccessPointID(strHost);
+                if (Character.isDigit(strHost.charAt(0))) {
+                    participant.setNetworkAccessPointTypeCode(AuditDataTransformConstants.NETWORK_ACCESSOR_PT_TYPE_CODE_IP);
+                } else {
+                    participant.setNetworkAccessPointTypeCode(CoreX12AuditDataTransformConstants.NETWORK_ACCESSOR_PT_TYPE_CODE_NAME);
+                }
+            } catch (MalformedURLException ex) {
+                LOG.error(ex);
```
jsmith2, line 205: We don't really handle this exception here.  I would take what's in the below else and pull out a private method and then call that from this catch if an exception happens.  that way the participant is set on error.

Product/Production/Services/CORE_X12DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/corex12/docsubmission/audit/transform/CoreX12AuditDataTransform.java
```diff
@@ -0,0 +1,467 @@
...
+        try {
+            JAXBContext jc = oHandler.getJAXBContext(CoreX12AuditDataTransformConstants.CORE_X12_JAXB_CONTEXT);
+            Marshaller marshaller = jc.createMarshaller();
+            ByteArrayOutputStream baOutStrm = new ByteArrayOutputStream();
+            if (msg instanceof COREEnvelopeRealTimeRequest) {
+                oRequestNoPayload = (COREEnvelopeRealTimeRequest) msg;
+                oRequestNoPayload.setPayload("");
+                xmlqname = new javax.xml.namespace.QName(CoreX12AuditDataTransformConstants.CORE_X12_NAMESPACE_URI, CoreX12AuditDataTransformConstants.CORE_X12_REQUEST_LOCALPART);
+                JAXBElement<org.caqh.soap.wsdl.corerule2_2_0.COREEnvelopeRealTimeRequest> element = new JAXBElement<org.caqh.soap.wsdl.corerule2_2_0.COREEnvelopeRealTimeRequest>(xmlqname, COREEnvelopeRealTimeRequest.class, oRequestNoPayload);
+                marshaller.marshal(element, baOutStrm);
```
jsmith2, line 279: these two bottom lines can be pulled from each if statement.  repetitive functionality.

Product/Production/Services/CORE_X12DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/corex12/docsubmission/audit/transform/CoreX12AuditDataTransform.java
```diff
@@ -0,0 +1,467 @@
...
+            LOG.error("response Object was null");
+            return null;
+        }
+        if (assertion == null) {
+            LOG.error("Assertion was null");
+            return null;
+        }
+        // check to see that the required fields are not null
+        boolean missingReqFields = CoreX12AuditDataTransform.areRequiredCOREX12BatchFiledsNull(response, assertion);
+        if (missingReqFields) {
```
jsmith2, line 414: Opening up a can of worms here, but I don't think this is right.  We fail to audit if any of these fields are missing?  I think we should be auditing what is there regardless.  So if user info is missing, for example, we fail to audit all the other information that is there?  I think we should discuss this.  Also that method that is used from XDR needs work.

PR #1071, opened 2015-05-23
Fhac 308- Fix to automate Direct Automated RegressionSuite

alameluchidambaram:
1. Two new params added to gateway.properties for Direct Testing. These params will be used only in case of Direct Testing.
2. MessageMonitoring Table data will be saved as "Archived" in MonitoredMessage Tables and those will be wiped off from the table only if the updateTime+delay time is before the CurrentTime. This accommodates Direct testing to hold values in db. The MessageMonitoring table Archived Messages won't be loaded into cache since they are already processed.


Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/messagemonitoring/impl/MessageMonitoringAPI.java
```diff
@@ -607,4 +628,60 @@ public void logErrorEvent(MimeMessage message, String errorMessage) {
         }
     }
 
+    private boolean getDirectTestFlag(String fileName, String property) {
+        boolean directTestFlag = false;
+        String directTestingFlag = getDirectTestingParam(fileName, property);
+        if (directTestingFlag.equals("true")) {
```
nareshcgi: Need to check if the directTestingFlag is null.

Product/Production/Common/Properties/src/main/resources/gateway.properties
```diff
@@ -130,6 +130,12 @@ DispatchedMessageReceiveTimeLimit=86400000
 AgentSettingsCacheRefreshTime=300000
 AgentSettingsCacheRefreshActive=true
 
+#Direct Testing parameters used only for Direct testing.
+#DirectTesting Flag accepts true/false value.
+#MessageMonitoringDelayTime accepts delay in minutes to wipe data from MessageMonitoringTable.
+DirectTesting=true
```
nareshcgi: Default value should be always false.

Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/messagemonitoring/impl/MessageMonitoringAPI.java
```diff
@@ -523,11 +538,17 @@ public void processAllMessages() throws MessageMonitoringDAOException {
...
+            //delete the message
+            if (getDirectTestFlag(NhincConstants.GATEWAY_PROPERTY_FILE, NhincConstants.DIRECTTESTING_FLAG)) {
+                trackMessage.setStatus(STATUS_ARCHIVED);
+                getMessageMonitoringDAO().updateOutgoingMessage(trackMessage);
+            } else {
+                deleteFromMessageMonitoringDB(trackMessage);
+            }
         }
+
+        deleteElapsedArchivedMessage();
```
nareshcgi: Make sure this method is called only when the direct testing flag is enabled.

Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/messagemonitoring/impl/MessageMonitoringAPI.java
```diff
@@ -245,8 +249,10 @@ private void buildCache() {
         clearCache();
         //load the pending outgoing messages to the cache
         for (MonitoredMessage trackMessage : pendingMessages) {
-            messageMonitoringCache.put(trackMessage.getMessageid(), trackMessage);
-            LOG.debug("Total child rows for the messageId:" + trackMessage.getMonitoredmessagenotifications().size());
+            if (!trackMessage.getStatus().equals(STATUS_ARCHIVED)) {
```
nareshcgi, line 252: Its better to cleanup the Archived message on a server restart. Call deleteElapsedArchivedMessage() method if we find any archived message.

nareshcgi:

PR #1072, opened 2015-05-28
FHAC-304: Code refactoring, cleanup and addressed previous PR comments

nareshcgi:
1. Done major code refactoring
2. Addressed previous PR comments
3. Added logic to set the client host address and the request URL
4. Removed unwanted code


Product/Production/Services/CORE_X12DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/corex12/docsubmission/audit/transform/CORE_X12AuditDataTransform.java
```diff
@@ -137,91 +204,78 @@ public static AuditSourceIdentificationType getAuditSourceIdentificationType() {
...
-    public static EventIdentificationType getEventIdentificationType(CodedValueType eventID, boolean isRequesting) {
+    protected EventIdentificationType getEventIdentificationType(CodedValueType eventID, boolean isRequesting) {
         EventIdentificationType oEventIdentificationType = AuditDataTransformHelper.createEventIdentification(
             isRequesting ? AuditDataTransformConstants.EVENT_ACTION_CODE_READ : AuditDataTransformConstants.EVENT_ACTION_CODE_CREATE,
             AuditDataTransformConstants.EVENT_OUTCOME_INDICATOR_SUCCESS, eventID);
         return oEventIdentificationType;
     }
 
-    public static AuditMessageType.ActiveParticipant getActiveParticipantSource(boolean isRequesting) {
+    protected AuditMessageType.ActiveParticipant getActiveParticipantSource(boolean isRequesting, Properties webContextProeprties) {
```
jsmith2: properties misspelled. webContextProperties

Product/Production/Services/CORE_X12DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/corex12/docsubmission/audit/transform/CORE_X12AuditDataTransform.java
```diff
@@ -379,89 +422,116 @@ private static boolean areRequiredRequestFieldsNull(Object msg) {
                 return true;
             } else if (NullChecker.isNullish(coreEnvelopeBatchSubmissionResponse.getTimeStamp())) {
                 LOG.error("CORE X12 Batch TimeStamp is empty...");
+
+                return true;
+            }
+        } else if (msg instanceof COREEnvelopeRealTimeRequest) {
+            coreEnvelopeRealTimeRequest = (COREEnvelopeRealTimeRequest) msg;
```
jsmith2: These are repetitive.  You could create a helper method that does something like areRequiredFieldsNull(String  ruleVersion, String payloadType, String processingMode ...., String messageType).  And just call that from each parent if statement.  messageType would be used for logging.

Product/Production/Services/CORE_X12DocumentSubmissionCore/pom.xml
```diff
@@ -51,6 +51,27 @@
             <groupId>commons-io</groupId>
             <artifactId>commons-io</artifactId>
         </dependency>
+        <!-- Testing -->
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.jmock</groupId>
```
jsmith2: Why was jmock added?  Don't see it used in your unit test.

nareshcgi: Those are the common libraries used across all the modules and these are only used for test, not packaged within the ear.

jsmith2: That's not a good reason to include them.  They're not being used and we've been using mockito instead of jmock as our preferred mocking library since release 4.0. Jmock is only there for older unit tests.

PR #1075, opened 2015-05-29
FHAC-293: Fixed the code to make AuthzDecisionStatement>Evidence>Assertion>Conditions SAML element not requried

nareshcgi:
Fixed the code to make AuthzDecisionStatement>Evidence>Assertion>Conditions SAML element not requried

### DIFF
--- a/Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/callback/openSAML/HOKSAMLAssertionBuilder.java
+++ b/Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/callback/openSAML/HOKSAMLAssertionBuilder.java
@@ -1,5 +1,28 @@
-/**
+/*
+ * Copyright (c) 2009-2015, United States Government, as represented by the Secretary of Health and Human Services.
+ * All rights reserved.
  *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *     * Redistributions of source code must retain the above
+ *       copyright notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the documentation
+ *       and/or other materials provided with the distribution.
+ *     * Neither the name of the United States Government nor the
+ *       names of its contributors may be used to endorse or promote products
+ *       derived from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE UNITED STATES GOVERNMENT BE LIABLE FOR ANY
+ * DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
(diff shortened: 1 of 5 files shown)


Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/callback/openSAML/HOKSAMLAssertionBuilder.java
```diff
@@ -690,4 +707,17 @@ public static Evidence buildEvidence(String evAssertionID, DateTime issueInstant
     private static String createAssertionId() {
         return ID_PREFIX.concat(String.valueOf(UUID.randomUUID())).replaceAll("-", "");
     }
+
+    protected boolean isAuthDEvidenceConditionsDefaultValueEnabled() {
+        //if not provided or invalid return true else false
+        try {
+            String authDEvidenceConditionsDefaultValueEnabled = PropertyAccessor.getInstance().getProperty(NhincConstants.GATEWAY_PROPERTY_FILE, NhincConstants.ENABLE_AUTH_DEC_EVIDENCE_CONDITIONS_DEFAULT_VALUE);
+            if (authDEvidenceConditionsDefaultValueEnabled != null) {
+                return !authDEvidenceConditionsDefaultValueEnabled.equals(Boolean.FALSE.toString());
```
jsmith2, line 716: this statement is confusing, shouldn't it just be: return Boolean.valueOf(authDEvidenceConditionDefaultValueEnabled);

that defaults to false if the value is not true.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/callback/openSAML/HOKSAMLAssertionBuilder.java
```diff
@@ -690,4 +707,17 @@ public static Evidence buildEvidence(String evAssertionID, DateTime issueInstant
...
+            }
+        } catch (PropertyAccessException pae) {
+            LOG.info("Proeprty Not found :" + NhincConstants.ENABLE_AUTH_DEC_EVIDENCE_CONDITIONS_DEFAULT_VALUE);
+        }
+        return Boolean.TRUE;
```
jsmith2, line 721: I think this should default to false not true.

nareshcgi, line 721: the default value should be true, since we want to retain the current CONNECT functionality.

Product/Production/Common/Properties/src/main/resources/gateway.properties
```diff
@@ -109,6 +109,12 @@ CoreX12GenericBatchTimeStampTimeToLive=300
 # This is only valid for  Core X12 Batch request and response services.
 CoreX12GenericBatchFutureTimeToLive=60
 
+# enableAuthDEvidenceConditionsDefaultValue accepts true or false, if not provided defaults to true.
```
jsmith2, line 112: Can we make this comment more concise.  Gonna be alot in the admin GUI display.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/callback/openSAML/HOKSAMLAssertionBuilder.java
nareshcgi, line 716: No this is also right
!authDEvidenceConditionsDefaultValueEnabled.equals(Boolean.FALSE.toString())
if the authDEvidenceConditionsDefaultValueEnabled is false then the above returns false
if its true, the above statement returns true
if its some other character, then the above statement returns true (default value)

Product/Production/Common/Properties/src/main/resources/gateway.properties
jsmith2, line 112: Example: "Boolean for adding the default values for AuthDecision Before and After Conditions.  Default property value is false."

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/callback/openSAML/HOKSAMLAssertionBuilder.java
jsmith2, line 716: Understand that it's logically right but the statement itself is confusing.  You're equating a double negative instead of just directly getting the value.

nareshcgi, line 716: but if i use Boolean.valueOf(authDEvidenceConditionDefaultValueEnabled), it will return false for values other than true, which is not right

jsmith2, line 716: I think the value in the property file should be true so noone has to do any updates when they upgrade but if someone accidentally changed it to property=crapValue, I think it's right to default to not use the default values.  We agreed that it's better to not use the values is best practice.  By setting it to true in the prop file it allows that no real upgrades are needed for the adopter that is upgrading.  Hopefully that makes sense.

nareshcgi:
Issues with the local branch, so created a new PR https://github.com/CONNECT-Solution/CONNECT/pull/1077 with the same code changes from this PR.

PR #1078, opened 2015-05-30
FHAC-308:Adding Delay step after Agent Settings Refresh for wsanchors-TestCase-7

alameluchidambaram:
Adding Delay test step to wsanchors-TestCase-7


nareshcgi:
QAd and looks good.

PR #1079, opened 2015-05-30
FHAC-308:Adding 5 minutes delay for Remote AgentSettings to be Refreshed

alameluchidambaram:
Adding 5 minutes delay for wsanchors-TestCase-7


nareshcgi:
PR looks good.

PR #1080, opened 2015-05-30
FHAC-296:Transaction Timeout Tests

sailajaa:
Added TransactionTimeout Tests
Removed return(space) between alias and name in x12 proxyconfig.xml files.


nareshcgi:
Any changes made to CORE_X12DSRealTimeProxyConfig.xml and CORE_X12DSGenericBatchProxyConfig.xml except formatting? If no, then revert back those files, unnecessary in this PR.

nareshcgi:
Not able to run the regression suite test..its hanging on the first Test Suite LargeTransactionTimeout --> Document Submission Mock Response.

sailajaa:
Removed retun(space) between alias and name in Core_x12RealTimeProxyConfig.xml and Core_X12DSGenericBatchProxyConfig.xml which was causing to update realtime adapter from secured to unsecred.

sailajaa:
Does it shows any error when its hanging at Mock step.  Ran LargeTransactionTimeout suite few times, could not able to reproduce error scenario.

jsmith2:
Transaction is spelled wrong in the project name.

sailajaa:
Renamed project name and updated testsuite description.

Product/Production/Common/Properties/src/main/resources/CORE_X12DSGenericBatchProxyConfig.xml
```diff
@@ -22,8 +22,7 @@ http://www.springframework.org/schema/aop     http://www.springframework.org/sch
     <!--
         Beans defined for adaptercore_x12dsgenericbatchrequest
     -->
-    <alias alias="adaptercore_x12dsgenericbatchrequest"
-           name="adaptercore_x12dsenericbatchrequestwsunsecured" />
+    <alias alias="adaptercore_x12dsgenericbatchrequest" name="adaptercore_x12dsenericbatchrequestwsunsecured" />
```
christophermay07, line 25: Is there any reason for changing the Proxy Config files?

christophermay07:
Why are AD test cases disabled by default?

christophermay07:
Following-up on Alamelu's comment about Passthrough:  If this test suite requires passthrough, what is the purpose of the "Check Passthru" test step in each test case?

christophermay07:
One last thing:  If you have to make another push, can you pretty-print save the SoapUI project file?

sailajaa:
AD test case can be removed from all three suites, since that is not used by anyone. But, idea to include and disable is, it might be helpful if we want to test in future if required.

christophermay07:

Product/Production/Common/Properties/src/main/resources/CORE_X12DSGenericBatchProxyConfig.xml
sailajaa, line 25: Removed retun(space) between alias and name in Core_x12RealTimeProxyConfig.xml and Core_X12DSGenericBatchProxyConfig.xml which was causing to update realtime adapter from secured to unsecred.

nareshcgi:
Looks good.

PR #1084, opened 2015-06-02
FHAC-320

christophermay07:
Updated SoapUI project files to pretty-printed XML (for ease of displaying diffs), and added a getPropertyValue fallback to global properties when all other sources are empty / null.


nareshcgi:

PR #1096, opened 2015-06-08
FHAC-368:Fix Mimemessage sender for Direct Outbound XDR Messages

alameluchidambaram:
Retrieve the sender from Mime Message "From" header when the MessageMonitoring is enabled for the Outbound XDR.

### DIFF
--- a/Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/MimeMessageBuilder.java
+++ b/Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/MimeMessageBuilder.java
@@ -141,7 +141,7 @@ public class MimeMessageBuilder {
 
         final MimeMessage message = new MimeMessage(session);
 
-        try {
+        try {    
             message.setFrom(fromAddress);
         } catch (Exception e) {
             throw new DirectException("Exception setting from address: " + fromAddress, e);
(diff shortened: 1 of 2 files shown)


Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/MimeMessageBuilder.java
```diff
@@ -141,7 +141,7 @@ public MimeMessage build() {
 
         final MimeMessage message = new MimeMessage(session);
 
-        try {
+        try {
```
christophermay07: Remove whitespace

Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/messagemonitoring/impl/MessageMonitoringAPI.java
```diff
@@ -694,4 +687,13 @@ private void deleteElapsedArchivedMessage(MonitoredMessage trackMessage) {
         }
     }
 
+    private String getSenderEmailId(MimeMessage message) throws MessagingException {
+        InternetAddress sender = (InternetAddress) message.getSender();
+        if (sender == null) {
+            InternetAddress[] fromAddresses = (InternetAddress[]) message.getFrom();
+            sender = fromAddresses[0];
```
christophermay07: Has "message" been null checked before this point?  Also, is there a guarantee that the fromAddresses length will be greater than 0?

alameluchidambaram: If the MimeMessage is "Null" it was already caught and exception was thrown by DirectSenderImpl sendOutboundDirect() method before reaching here .

christophermay07:

nareshcgi:
Good with the PR.

PR #1098, opened 2015-06-09
FHAC-386:Fix Direct suite testcases for emailId,endpoint and wsdl path

alameluchidambaram:
1. Fixed the endpoint for soapui requests
2. Corrected "From" and "To" valid email Id for soapui requests.
3. Fix Project Dir and wsdl path


nareshcgi:

PR #1102, opened 2015-06-26
FHAC-420: Added null check in the CXFClient to resolve Null Pointer Exception issue

nareshcgi:
Added null check in the CXFClient to resolve Null Pointer Exception issue


alameluchidambaram:
Reviewed and good with merge.

PR #1111, opened 2015-07-28
FHAC-463: Updated complier source, target values to 1.7 in POM

sailajaa:
1. Updated complier souce, targer values to 1.7  in Connect/pom.xml and connect/product/pom.xml


christophermay07:
I've reviewed this PR, endorsing for merge without comment.

PR #1112, opened 2015-07-30
FHAC 456 - Separate tests in automated Direct suite skippable

sovanncgi:
With this update, parameters can be passed in from the nightly CI to determine which automated Direct test cases should be run and which should be skipped. Delay time before performing event assertions has been increased from 300000 to 350000. This gives us more flexibility for testing and validating Direct using nightly CI.


nareshcgi:

PR #1116, opened 2015-08-18
FHAC-480 Added testcases for PD auditlog xml attribute Verify

sailajaa:
1. Added Testcases to verify PD auditlog xml message for standard and passthrough


sailajaa:
AuditLogging suite already have X12 project for which properties are defined and it will be loaded when the project opens up. But the suggestion I got when I created this suite was to maintain individual properties files for each service, so for PD I created seperate properties file.

PR #1117, opened 2015-08-19
Fhac 487 - ATNA complaint PatientDiscovery Audit Logger

alameluchidambaram:
AuditTransformer is created in AuditRepositoryCore which can be utilized across services for Audit Logging.  PatientDiscovery Transformer is also created to transform objects specific to PatientDiscovery request and responses. Patient Discovery Audit Logger is created and calls were modified in Outbound and inbound. The number of audit log entries remain the same both in Standard and Passthrough mode.

### DIFF
--- a/Product/Production/Gateway/PatientDiscovery_10/src/test/java/gov/hhs/fha/nhinc/patientdiscovery/_10/gateway/ws/PatientDiscoverySpringContextTest.java
+++ b/Product/Production/Gateway/PatientDiscovery_10/src/test/java/gov/hhs/fha/nhinc/patientdiscovery/_10/gateway/ws/PatientDiscoverySpringContextTest.java
@@ -26,14 +26,8 @@
  */
 package gov.hhs.fha.nhinc.patientdiscovery._10.gateway.ws;
 
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertTrue;
-import static org.mockito.Matchers.any;
-import static org.mockito.Matchers.eq;
-import static org.mockito.Mockito.mock;
-import static org.mockito.Mockito.when;
 import gov.hhs.fha.nhinc.common.nhinccommon.AssertionType;
+import gov.hhs.fha.nhinc.nhinclib.NhincConstants;
 import gov.hhs.fha.nhinc.patientdiscovery.PatientDiscoveryException;
 import gov.hhs.fha.nhinc.patientdiscovery.inbound.InboundPatientDiscovery;
 import gov.hhs.fha.nhinc.patientdiscovery.inbound.PassthroughInboundPatientDiscovery;
@@ -41,25 +35,48 @@ import gov.hhs.fha.nhinc.patientdiscovery.inbound.StandardInboundPatientDiscover
 import gov.hhs.fha.nhinc.patientdiscovery.outbound.PassthroughOutboundPatientDiscovery;
 import gov.hhs.fha.nhinc.patientdiscovery.outbound.StandardOutboundPatientDiscovery;
 import ihe.iti.xcpd._2009.PRPAIN201305UV02Fault;
-
+import java.security.Principal;
+import java.util.Properties;
+import javax.annotation.PostConstruct;
+import javax.servlet.http.HttpServletRequest;
+import javax.xml.ws.EndpointReference;
+import javax.xml.ws.WebServiceContext;
+import javax.xml.ws.handler.MessageContext;
+import org.apache.cxf.jaxws.context.WebServiceContextImpl;
+import org.apache.cxf.message.Message;
+import org.apache.cxf.transport.http.AbstractHTTPDestination;
 import org.hl7.v3.PRPAIN201305UV02;
 import org.hl7.v3.PRPAIN201306UV02;
 import org.hl7.v3.RespondingGatewayPRPAIN201305UV02RequestType;
 import org.hl7.v3.RespondingGatewayPRPAIN201306UV02ResponseType;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertTrue;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 import org.junit.runner.RunWith;
+import static org.mockito.Matchers.any;
+import static org.mockito.Matchers.eq;
+import org.mockito.Mockito;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.when;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.test.context.ContextConfiguration;
 import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+import org.w3c.dom.Element;
 
 /**
  * @author akong
  *
  */
 @RunWith(SpringJUnit4ClassRunner.class)
-@ContextConfiguration(locations = { "/patientdiscovery/_10/applicationContext.xml" })
+@ContextConfiguration(locations = {"/patientdiscovery/_10/applicationContext.xml"})
 public class PatientDiscoverySpringContextTest {
 
     @Rule
@@ -89,44 +106,43 @@ public class PatientDiscoverySpringContextTest {
     @Test
     public void inbound() throws PRPAIN201305UV02Fault {
         assertNotNull(inboundPatientDiscoveryEndpoint);
-
         PRPAIN201305UV02 request = new PRPAIN201305UV02();
+        WebServiceContext context = new WebServiceContextImpl();
+        inboundPatientDiscoveryEndpoint.setContext(context);
         PRPAIN201306UV02 response = inboundPatientDiscoveryEndpoint.respondingGatewayPRPAIN201305UV02(request);
-
         assertNotNull(response);
     }
 
-   /* @Test
-    public void inboundFault() throws PatientDiscoveryException {
-        PRPAIN201305UV02 request = new PRPAIN201305UV02();
+    /* @Test
+     public void inboundFault() throws PatientDiscoveryException {
+     PRPAIN201305UV02 request = new PRPAIN201305UV02();
 
-        NhinPatientDiscovery inboundPDEndpoint = new NhinPatientDiscovery();
+     NhinPatientDiscovery inboundPDEndpoint = new NhinPatientDiscovery();
 
-        InboundPatientDiscovery inboundPatientDiscovery = mock(InboundPatientDiscovery.class);
+     InboundPatientDiscovery inboundPatientDiscovery = mock(InboundPatientDiscovery.class);
 
-        when(inboundPatientDiscovery.respondingGatewayPRPAIN201305UV02(eq(request), any(AssertionType.class)))
-                .thenThrow(new PatientDiscoveryException(""));
+     when(inboundPatientDiscovery.respondingGatewayPRPAIN201305UV02(eq(request), any(AssertionType.class)))
+     .thenThrow(new PatientDiscoveryException(""));
 
-        inboundPDEndpoint.setInboundPatientDiscovery(inboundPatientDiscovery);
+     inboundPDEndpoint.setInboundPatientDiscovery(inboundPatientDiscovery);
 
-        boolean faultThrown = false;
-        try {
-            inboundPDEndpoint.respondingGatewayPRPAIN201305UV02(request);
-        } catch (PRPAIN201305UV02Fault fault) {
-            faultThrown = true;
-            assertEquals("920", fault.getFaultInfo().getErrorCode());
-        }
+     boolean faultThrown = false;
+     try {
+     inboundPDEndpoint.respondingGatewayPRPAIN201305UV02(request);
+     } catch (PRPAIN201305UV02Fault fault) {
+     faultThrown = true;
+     assertEquals("920", fault.getFaultInfo().getErrorCode());
+     }
 
-        assertTrue(faultThrown);
-    }*/
-
+     assertTrue(faultThrown);
+     }*/
     @Test
     public void outboundUnsecured() {
         assertNotNull(outboundPatientDiscoveryUnsecuredEndpoint);
 
         RespondingGatewayPRPAIN201305UV02RequestType request = new RespondingGatewayPRPAIN201305UV02RequestType();
         RespondingGatewayPRPAIN201306UV02ResponseType response = outboundPatientDiscoveryUnsecuredEndpoint
-                .respondingGatewayPRPAIN201305UV02(request);
+            .respondingGatewayPRPAIN201305UV02(request);
 
         assertNotNull(response);
     }
@@ -137,7 +153,7 @@ public class PatientDiscoverySpringContextTest {
 
         RespondingGatewayPRPAIN201305UV02RequestType request = new RespondingGatewayPRPAIN201305UV02RequestType();
         RespondingGatewayPRPAIN201306UV02ResponseType response = outboundPatientDiscoverySecuredEndpoint
-                .respondingGatewayPRPAIN201305UV02(request);
+            .respondingGatewayPRPAIN201305UV02(request);
 
         assertNotNull(response);
     }
--- a/Product/Production/Gateway/PatientDiscovery_10/src/test/java/gov/hhs/fha/nhinc/patientdiscovery/_10/gateway/ws/PatientDiscoverySpringContextTest.java
+++ b/Product/Production/Gateway/PatientDiscovery_10/src/test/java/gov/hhs/fha/nhinc/patientdiscovery/_10/gateway/ws/PatientDiscoverySpringContextTest.java
@@ -27,7 +27,6 @@
 package gov.hhs.fha.nhinc.patientdiscovery._10.gateway.ws;
 
 import gov.hhs.fha.nhinc.common.nhinccommon.AssertionType;
-import gov.hhs.fha.nhinc.nhinclib.NhincConstants;
 import gov.hhs.fha.nhinc.patientdiscovery.PatientDiscoveryException;
 import gov.hhs.fha.nhinc.patientdiscovery.inbound.InboundPatientDiscovery;
 import gov.hhs.fha.nhinc.patientdiscovery.inbound.PassthroughInboundPatientDiscovery;
@@ -35,41 +34,28 @@ import gov.hhs.fha.nhinc.patientdiscovery.inbound.StandardInboundPatientDiscover
 import gov.hhs.fha.nhinc.patientdiscovery.outbound.PassthroughOutboundPatientDiscovery;
 import gov.hhs.fha.nhinc.patientdiscovery.outbound.StandardOutboundPatientDiscovery;
 import ihe.iti.xcpd._2009.PRPAIN201305UV02Fault;
-import java.security.Principal;
 import java.util.Properties;
-import javax.annotation.PostConstruct;
-import javax.servlet.http.HttpServletRequest;
-import javax.xml.ws.EndpointReference;
 import javax.xml.ws.WebServiceContext;
-import javax.xml.ws.handler.MessageContext;
 import org.apache.cxf.jaxws.context.WebServiceContextImpl;
-import org.apache.cxf.message.Message;
-import org.apache.cxf.transport.http.AbstractHTTPDestination;
 import org.hl7.v3.PRPAIN201305UV02;
 import org.hl7.v3.PRPAIN201306UV02;
 import org.hl7.v3.RespondingGatewayPRPAIN201305UV02RequestType;
 import org.hl7.v3.RespondingGatewayPRPAIN201306UV02ResponseType;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
+import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 import org.junit.runner.RunWith;
 import static org.mockito.Matchers.any;
 import static org.mockito.Matchers.eq;
-import org.mockito.Mockito;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.test.context.ContextConfiguration;
 import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
-import org.w3c.dom.Element;
 
 /**
  * @author akong
@@ -113,7 +99,8 @@ public class PatientDiscoverySpringContextTest {
         assertNotNull(response);
     }
 
-    /* @Test
+    @Ignore 
+    @Test
      public void inboundFault() throws PatientDiscoveryException {
      PRPAIN201305UV02 request = new PRPAIN201305UV02();
 
@@ -121,7 +108,8 @@ public class PatientDiscoverySpringContextTest {
 
      InboundPatientDiscovery inboundPatientDiscovery = mock(InboundPatientDiscovery.class);
 
-     when(inboundPatientDiscovery.respondingGatewayPRPAIN201305UV02(eq(request), any(AssertionType.class)))
+     when(inboundPatientDiscovery.respondingGatewayPRPAIN201305UV02(eq(request), any(AssertionType.class), 
+         any(Properties.class)))
      .thenThrow(new PatientDiscoveryException(""));
 
      inboundPDEndpoint.setInboundPatientDiscovery(inboundPatientDiscovery);
@@ -135,7 +123,7 @@ public class PatientDiscoverySpringContextTest {
      }
 
      assertTrue(faultThrown);
-     }*/
+     }
     @Test
     public void outboundUnsecured() {
         assertNotNull(outboundPatientDiscoveryUnsecuredEndpoint);
--- a/Product/Production/Gateway/PatientDiscovery_10/src/test/java/gov/hhs/fha/nhinc/patientdiscovery/_10/gateway/ws/PatientDiscoverySpringContextTest.java
+++ b/Product/Production/Gateway/PatientDiscovery_10/src/test/java/gov/hhs/fha/nhinc/patientdiscovery/_10/gateway/ws/PatientDiscoverySpringContextTest.java
@@ -99,31 +99,32 @@ public class PatientDiscoverySpringContextTest {
         assertNotNull(response);
     }
 
-    @Ignore 
+    @Ignore
     @Test
-     public void inboundFault() throws PatientDiscoveryException {
-     PRPAIN201305UV02 request = new PRPAIN201305UV02();
+    public void inboundFault() throws PatientDiscoveryException {
+        PRPAIN201305UV02 request = new PRPAIN201305UV02();
 
-     NhinPatientDiscovery inboundPDEndpoint = new NhinPatientDiscovery();
+        NhinPatientDiscovery inboundPDEndpoint = new NhinPatientDiscovery();
 
-     InboundPatientDiscovery inboundPatientDiscovery = mock(InboundPatientDiscovery.class);
+        InboundPatientDiscovery inboundPatientDiscovery = mock(InboundPatientDiscovery.class);
 
-     when(inboundPatientDiscovery.respondingGatewayPRPAIN201305UV02(eq(request), any(AssertionType.class), 
-         any(Properties.class)))
-     .thenThrow(new PatientDiscoveryException(""));
+        when(inboundPatientDiscovery.respondingGatewayPRPAIN201305UV02(eq(request), any(AssertionType.class),
+            any(Properties.class)))
+            .thenThrow(new PatientDiscoveryException(""));
 
-     inboundPDEndpoint.setInboundPatientDiscovery(inboundPatientDiscovery);
+        inboundPDEndpoint.setInboundPatientDiscovery(inboundPatientDiscovery);
 
-     boolean faultThrown = false;
-     try {
-     inboundPDEndpoint.respondingGatewayPRPAIN201305UV02(request);
-     } catch (PRPAIN201305UV02Fault fault) {
-     faultThrown = true;
-     assertEquals("920", fault.getFaultInfo().getErrorCode());
-     }
+        boolean faultThrown = false;
+        try {
+            inboundPDEndpoint.respondingGatewayPRPAIN201305UV02(request);
+        } catch (PRPAIN201305UV02Fault fault) {
+            faultThrown = true;
+            assertEquals("920", fault.getFaultInfo().getErrorCode());
+        }
 
-     assertTrue(faultThrown);
-     }
+        assertTrue(faultThrown);
+    }
+
     @Test
     public void outboundUnsecured() {
         assertNotNull(outboundPatientDiscoveryUnsecuredEndpoint);
--- a/Product/Production/Services/AuditRepositoryCore/pom.xml
+++ b/Product/Production/Services/AuditRepositoryCore/pom.xml
@@ -51,4 +51,20 @@
             <artifactId>junit</artifactId>
         </dependency>
     </dependencies>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <version>2.6</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>test-jar</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
 </project>
--- a/Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/AuditTransformConstants.java
+++ b/Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/AuditTransformConstants.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright (c) 2009-2015, United States Government, as represented by the Secretary of Health and Human Services.
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *     * Redistributions of source code must retain the above
+ *       copyright notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the documentation
+ *       and/or other materials provided with the distribution.
+ *     * Neither the name of the United States Government nor the
+ *       names of its contributors may be used to endorse or promote products
+ *       derived from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ * ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+ * DISCLAIMED. IN NO EVENT SHALL THE UNITED STATES GOVERNMENT BE LIABLE FOR ANY
+ * DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+ * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
+ * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+package gov.hhs.fha.nhinc.audit;
+
+/**
+ *
+ * @author achidamb
+ */
+public class AuditTransformConstants {
+    public static final Integer EVENT_OUTCOME_INDICATOR_SUCCESS = 0;
+    public static final String ACTIVE_PARTICPANT_USER_ID_SOURCE = "anonymous";
+    public static final String ACTIVE_PARTICPANT_ROLE_CODE_SOURCE_DISPLAY_NAME = "Source";
+    public static final String ACTIVE_PARTICPANT_ROLE_CODE_DESTINATION_DISPLAY_NAME = "Destination";
+    public static final String ACTIVE_PARTICPANT_UNKNOWN_IP_ADDRESS = "unknown";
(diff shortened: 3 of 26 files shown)


Product/Production/Gateway/PatientDiscovery_10/src/test/java/gov/hhs/fha/nhinc/patientdiscovery/_10/gateway/ws/PatientDiscoverySpringContextTest.java
```diff
@@ -89,44 +106,43 @@
...
-
         assertNotNull(response);
     }
 
-   /* @Test
-    public void inboundFault() throws PatientDiscoveryException {
-        PRPAIN201305UV02 request = new PRPAIN201305UV02();
-
-        NhinPatientDiscovery inboundPDEndpoint = new NhinPatientDiscovery();
+    /* @Test
```
christophermay07: If the test will compile, use the "Ignore" annotation instead of commenting it out.

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/AuditTransformConstants.java
```diff
@@ -0,0 +1,45 @@
...
+ */
+public class AuditTransformConstants {
+    public static final Integer EVENT_OUTCOME_INDICATOR_SUCCESS = 0;
+    public static final String ACTIVE_PARTICPANT_USER_ID_SOURCE = "anonymous";
+    public static final String ACTIVE_PARTICPANT_ROLE_CODE_SOURCE_DISPLAY_NAME = "Source";
+    public static final String ACTIVE_PARTICPANT_ROLE_CODE_DESTINATION_DISPLAY_NAME = "Destination";
+    public static final String ACTIVE_PARTICPANT_UNKNOWN_IP_ADDRESS = "unknown";
+    public static final Short NETWORK_ACCESSOR_PT_TYPE_CODE_NAME = 2;
+    public static final String ACTIVE_PARTICIPANT_ROLE_CODE_DEST = "110152";
+    public static final String ACTIVE_PARTICIPANT_ROLE_CODE_Source = "110153";
```
christophermay07: Should be "SOURCE" instead of "Source"

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/AuditTransformDataBuilder.java
```diff
@@ -0,0 +1,194 @@
...
+     */
+    public String getEventIdCode(String serviceName) {
+        if (eventIdCodeMap.containsKey(serviceName)) {
+            return eventIdCodeMap.get(serviceName);
+        }
+        return null;
+    }
+
+    private Map<String, String> createEventCodeSystemMap() {
+        eventCodeSystemMap = new HashMap();
```
christophermay07: Since we're using JDK 7 now, use the diamond operator **EVERYWHERE** you invoke "new" on a type that accepts generics, e.g.:

new HashMap<>();
new ArrayList<>();
etc.

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/AuditTransformDataBuilder.java
```diff
@@ -0,0 +1,194 @@
...
+    /**
+     *
+     * @param serviceName
+     * @return
+     */
+    public String getEventIdCode(String serviceName) {
+        if (eventIdCodeMap.containsKey(serviceName)) {
+            return eventIdCodeMap.get(serviceName);
+        }
+        return null;
```
christophermay07: HashMap's "get" returns null if it can't find the key; see http://docs.oracle.com/javase/7/docs/api/java/util/HashMap.html#get(java.lang.Object)

Everywhere you have a function like this you can simplify to a single line, in this case:

return eventIdCodeMap.get(serviceName);

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/AuditTransformDataBuilder.java
```diff
@@ -0,0 +1,194 @@
...
+        eventTypeCodeSystemMap = createEventTypeCodeSystemMap();
+        eventTypeCodeDisplayNameMap = createEventTypeCodeDisplayNameMap();
+        eventActionCodeRequestorMap = createEventActionCodeRequestorMap();
+        eventActionCodeResponderMap = createEventActionCodeResponderMap();
+    }
+
+    private Map<String, String> createEventIdCodeMap() {
+        eventIdCodeMap = new HashMap();
+        eventIdCodeMap.put("PatientDiscovery", "110112");
+        return eventIdCodeMap;
```
christophermay07: This logic is repeated multiple times; consider using a private helper function like "createSingletonStringMap(String, String)", and using that repeatedly, e.g.:

return createSingletonStringMap("PatientDiscovery", "110112");

This way, you're not repeating the map creation and insertion.

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/transform/AuditTransform.java
```diff
@@ -0,0 +1,579 @@
...
+        result.setDirection(direction);
+        result.setInterface(_interface);
+        //set the target community identifier
+        result.setCommunityId(communityId);
+
+        LOG.trace("End AuditDataTransform transformMsgToAuditMsg() ----");
+        return result;
+    }
+
+    /**
```
christophermay07, line 141: As a rule of thumb, we want full JavaDoc comments everywhere we add / modify code, but it's **extremely** important for abstract methods in template classes -- this will make it easier if someone else ends up writing an implementation later.

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/transform/AuditTransform.java
```diff
@@ -0,0 +1,579 @@
...
+     * @param eventID
+     * @param isRequesting
+     * @param serviceName
+     * @return
+     */
+    protected EventIdentificationType getEventIdentificationType(CodedValueType eventID, boolean isRequesting, String serviceName) {
+        EventIdentificationType oEventIdentificationType = createEventIdentification(
+            isRequesting ? getEventActionCodeRequestor(serviceName) : getEventActionCodeResponder(serviceName),
+            AuditTransformConstants.EVENT_OUTCOME_INDICATOR_SUCCESS, eventID);
+        return oEventIdentificationType;
```
christophermay07: Don't create a placeholder object just to immediately return it; change this to an inline return.

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/transform/AuditTransform.java
```diff
@@ -0,0 +1,579 @@
...
+                participant.setNetworkAccessPointID(strHost);
+                participant.setNetworkAccessPointTypeCode(getNetworkAccessPointTypeCode(strHost));
+                setDefaultValue = false;
+            } catch (MalformedURLException ex) {
+                LOG.error(ex);
+                setDefaultValue = true;
+            }
+        }
+        //if the url is null or not a valid url
+        if (setDefaultValue) {
```
christophermay07: Can this logic be added in the catch, this eliminating the need for both this "if" statement as well as the "setDefaultValue" variable?

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/transform/AuditTransform.java
```diff
@@ -0,0 +1,579 @@
...
+    }
+
+    /**
+     *
+     * @param webContextProeprties
+     * @return
+     */
+    protected String getRemoteHostAddress(Properties webContextProeprties) {
+        if (webContextProeprties != null && !webContextProeprties.isEmpty() && webContextProeprties.getProperty(NhincConstants.REMOTE_HOST_ADDRESS) != null) {
+            return webContextProeprties.getProperty(NhincConstants.REMOTE_HOST_ADDRESS);
```
christophermay07: Typo, should be "webContextProperties" -- do a search and replace, this happens at least one other time in an unrelated place.

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/transform/AuditTransform.java
```diff
@@ -0,0 +1,579 @@
...
+
+    /**
+     *
+     * @param hostAddress
+     * @return
+     */
+    protected Short getNetworkAccessPointTypeCode(String hostAddress) {
+        if (InetAddressValidator.getInstance().isValid(hostAddress)) {
+            return AuditTransformConstants.NETWORK_ACCESSOR_PT_TYPE_CODE_IP;
+        } else {
```
christophermay07: This is the only place you put an "else" after the "if" returns -- for consistency, change this to not use an else statement.

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/transform/AuditTransform.java
```diff
@@ -0,0 +1,579 @@
...
+        EventIdentificationType eventIdentification = new EventIdentificationType();
+
+        // Set the Event Action Code
+        eventIdentification.setEventActionCode(actionCode);
+
+        // Set the Event Action Time
+        try {
+            java.util.GregorianCalendar today = new java.util.GregorianCalendar(TimeZone.getTimeZone("GMT"));
+            javax.xml.datatype.DatatypeFactory factory = javax.xml.datatype.DatatypeFactory.newInstance();
+            javax.xml.datatype.XMLGregorianCalendar calendar = factory.newXMLGregorianCalendar(
```
christophermay07: Why are you using fully-qualified types for the previous three lines?

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/transform/AuditTransform.java
```diff
@@ -0,0 +1,579 @@
...
+                today.get(java.util.GregorianCalendar.YEAR), today.get(java.util.GregorianCalendar.MONTH) + 1,
+                today.get(java.util.GregorianCalendar.DAY_OF_MONTH),
+                today.get(java.util.GregorianCalendar.HOUR_OF_DAY), today.get(java.util.GregorianCalendar.MINUTE),
+                today.get(java.util.GregorianCalendar.SECOND), today.get(java.util.GregorianCalendar.MILLISECOND),
+                0);
+            eventIdentification.setEventDateTime(calendar);
+        } catch (DatatypeConfigurationException e) {
+            LOG.error("DatatypeConfigurationException when creating XMLGregorian Date");
+            LOG.error(" message: " + e.getMessage());
```
christophermay07: Anywhere you log an exception, using the following format: LOG.error("<message>" + e.getLocalizedMessage(), e);

For example, in this case, try:

LOG.error("Exception when creating XMLGregorian Date" + e.getLocalizedMessage(), e);

Note that you don't need to explicitly list "DatatypeConfigurationException"; the stack trace will take care of that.

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/transform/AuditTransform.java
```diff
@@ -0,0 +1,579 @@
...
+            return AuditTransformConstants.ACTIVE_PARTICPANT_UNKNOWN_IP_ADDRESS;
+        }
+    }
+
+    /**
+     *
+     * @return
+     */
+    protected String createUUID() {
+        return (UUID.randomUUID().toString());
```
christophermay07: Unnecessary parentheses (and an extra line break).

Product/Production/Services/PatientDiscoveryCore/src/main/java/gov/hhs/fha/nhinc/patientdiscovery/audit/transform/PatientDiscoveryTransforms.java
```diff
@@ -0,0 +1,187 @@
...
+     * @param response
+     * @return
+     */
+    @Override
+    protected AuditMessageType getParticipantObjectIdentification(PRPAIN201305UV02 request, PRPAIN201306UV02 response, AssertionType assertion, AuditMessageType auditMsg) {
+        auditMsg = getPatientParticipantObjectIdentification(request, response, auditMsg);
+        try {
+            auditMsg = getQueryParamsParticipantObjectIdentification(request, response, auditMsg);
+        } catch (JAXBException ex) {
+            LOG.debug("Error while creating ParticipantObjectIdentificationQueryByParameters segemnt : " + ex);
```
christophermay07: Typo, should be "segment" -- also, change this to LOG.error if it's an error, or change to LOG.warn("Unable to create ...." ... )

Also, since I'm already commenting here, just a reminder -- EVERYWHERE you log an exception, use ("msg" + ex.getLocalizedMessage(), ex)

Product/Production/Services/PatientDiscoveryCore/src/main/java/gov/hhs/fha/nhinc/patientdiscovery/audit/transform/PatientDiscoveryTransforms.java
```diff
@@ -0,0 +1,187 @@
...
+            && oPatientDiscoveryRequestMessage.getControlActProcess().getQueryByParameter().getValue() != null
+            && oPatientDiscoveryRequestMessage.getControlActProcess().getQueryByParameter().getValue().getParameterList() != null
+            && oPatientDiscoveryRequestMessage.getControlActProcess().getQueryByParameter().getValue().getParameterList().getLivingSubjectId() != null
+            && oPatientDiscoveryRequestMessage.getControlActProcess().getQueryByParameter().getValue().getParameterList().getLivingSubjectId().get(0) != null && oPatientDiscoveryRequestMessage.getControlActProcess().getQueryByParameter().getValue().getParameterList().getLivingSubjectId().get(0).getValue() != null && oPatientDiscoveryRequestMessage.getControlActProcess().getQueryByParameter().getValue().getParameterList().getLivingSubjectId().get(0).getValue().get(0) != null) {
+            oII = oPatientDiscoveryRequestMessage.getControlActProcess().getQueryByParameter().getValue().getParameterList().getLivingSubjectId().get(0).getValue().get(0);
+
+        } else {
+            LOG.error("PatientId doesn't exists in the received PRPAIN201305UV02 message");
+            return oII;
```
christophermay07: Unnecessary return (you immediately return outside of the else).

Product/Production/Services/PatientDiscoveryCore/src/main/java/gov/hhs/fha/nhinc/patientdiscovery/audit/transform/PatientDiscoveryTransforms.java
```diff
@@ -0,0 +1,187 @@
...
+        } else {
+            auditMsg.getParticipantObjectIdentification().add(participantObject);
+        }
+        return auditMsg;
+    }
+
+    private static String createPatientId(String assigningAuthId, String patientId) {
+        String sValue = null;
+        sValue = patientId + "^^^&" + assigningAuthId + "&ISO";
+        return sValue;
```
christophermay07: Condense this to a single line.

Product/Production/Services/PatientDiscoveryCore/src/main/java/gov/hhs/fha/nhinc/patientdiscovery/inbound/PassthroughInboundPatientDiscovery.java
```diff
@@ -78,12 +77,12 @@ PRPAIN201306UV02 process(PRPAIN201305UV02 body, AssertionType assertion) throws
...
-    PatientDiscoveryAuditor getAuditLogger() {
+    PatientDiscoveryAuditLogger getAuditLogger() {
         return auditLogger;
     }
 
     private PRPAIN201306UV02 sendToAdapter(PRPAIN201305UV02 request, AssertionType assertion)
-            throws PatientDiscoveryException {
+        throws PatientDiscoveryException {
         AdapterPatientDiscoveryProxy proxy = adapterFactory.create();
         return proxy.respondingGatewayPRPAIN201305UV02(request, assertion);
```
christophermay07: Simplify this to one line.

Product/Production/Services/PatientDiscoveryCore/src/test/java/gov/hhs/fha/nhinc/patientdiscovery/inbound/PassthroughInboundPatientDiscoveryTest.java
```diff
@@ -26,25 +26,28 @@
...
 import gov.hhs.fha.nhinc.patientdiscovery.PatientDiscoveryException;
 import gov.hhs.fha.nhinc.patientdiscovery.adapter.proxy.AdapterPatientDiscoveryProxy;
 import gov.hhs.fha.nhinc.patientdiscovery.adapter.proxy.AdapterPatientDiscoveryProxyObjectFactory;
-
+import gov.hhs.fha.nhinc.patientdiscovery.audit.PatientDiscoveryAuditLogger;
+import java.util.Properties;
 import org.hl7.v3.PRPAIN201305UV02;
 import org.hl7.v3.PRPAIN201306UV02;
+import static org.junit.Assert.assertSame;
+import static org.junit.Assert.assertSame;
```
christophermay07: You have a duplicated import here, and another down below.

Product/Production/Services/AuditRepositoryCore/pom.xml
```diff
@@ -51,4 +51,20 @@
...
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <version>2.6</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>test-jar</goal>
```
christophermay07, line 63: I've seen this test-jar goal in a few places before -- what is it used for?

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/transform/AuditTransform.java
```diff
@@ -0,0 +1,579 @@
...
+    /**
+     *
+     * @param oUserInfo
+     * @return
+     */
+    protected AuditMessageType.ActiveParticipant getActiveParticipant(UserType oUserInfo) {
+        // Create Active Participant Section
+        // create a method to call the AuditDataTransformHelper - one expectation
+        AuditMessageType.ActiveParticipant participant = createActiveParticipantFromUser(
+            oUserInfo, Boolean.TRUE);
```
christophermay07: You line wrap MUCH later a couple lines down; should this wrap later on this line, or wrapped earlier down below?

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/transform/AuditTransform.java
```diff
@@ -0,0 +1,579 @@
...
+
+    /**
+     *
+     * @param serviceName
+     * @param isRequesting
+     * @return
+     */
+    protected EventIdentificationType createEventIdentification(String serviceName, boolean isRequesting) {
+        CodedValueType eventID = createCodeValueType(getEventIdCode(serviceName), null,
+            getEventCodeSystemName(serviceName), isRequesting ? getEventDisplayNameRequestor(serviceName) : getEventDisplayNameResponder(serviceName));
```
christophermay07: As before, make sure you're line wrapping at or before 120 characters.

Product/Production/Services/AuditRepositoryCore/pom.xml
alameluchidambaram, line 63: The AuditTransform abstract class tests are written in AuditRepositoryCore services jar. Since these are same across all services PD,DQ,DR,DS,AD,X12 and in order to avoid repetition of writing unit tests for same code an abstract AuditTransform testclass is created and the AuditRepositoryCore  test jar is added as dependency to PatientDiscovery Core services. Same way it can be used in other Core services and the service specific AuditTransformer uni tests can be written in their respective services core.

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/AuditTransformConstants.java
jsmith2: Is there any value in moving these constants to a property file?

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/AuditTransformDataBuilder.java
jsmith2: I don't think this is the right way to do this.  I think you probably should create a hashmap per service and put that map in the service's lib.

for example, you'd have a class in PD service jar that has the map:

Map<String,String> patientDiscoveryAuditMap = new HashMap<>():

private void loadMap() {
     patientDiscoveryAuditMap.put("eventDisplayName", "PatientDiscovery");
     patientDiscoveryAuditMap.put("eventTypeCode", "ITI-55");
     ...
}

jsmith2: The main reason for the above suggestion is that you're putting service specific info into the auditing library, seems to be the wrong spot.  These values also could be stored in a property file.

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/transform/AuditTransform.java
```diff
@@ -0,0 +1,579 @@
...
+     * @param isRequesting
+     * @param webContextProperties
+     * @param serviceName
+     * @param response
+     * @return
+     */
+    public final LogEventRequestType transformMsgToAuditMsg(T request, AssertionType assertion,
+        NhinTargetSystemType target, String direction, String _interface, boolean isRequesting, Properties webContextProperties, String serviceName, K response) {
+        LOG.trace("Begin AuditDataTransform -> transformMsgToAuditMsg() ----");
+        if (request == null && response == null) {
```
jsmith2: Should this be an or?  Should we audit if there is no response but there is a request?

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/transform/AuditTransform.java
jsmith2: Don't think we need the boolean as a method argument if it's always "TRUE".  Can just set it in the private method.

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/transform/AuditTransform.java
```diff
@@ -0,0 +1,579 @@
...
+    /**
+     * Create the <code>CodedValueType</code> for an audit log record.
+     *
+     * @param code
+     * @param codeSys
+     * @param codeSysName
+     * @param dispName
+     * @return <code>CodedValueType</code>
+     */
+    public static CodedValueType createCodeValueType(String code, String codeSys, String codeSysName, String dispName) {
```
jsmith2: Should this method be private?  Is it used outside the class?

Product/Production/Services/AuditRepositoryCore/src/test/java/gov/hhs/fha/nhinc/audit/transform/AuditTransformTest.java
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    @After
+    public void tearDown() {
+    }
+
+    /**
+     * Test of transformMsgToAuditMsg method, of class AuditTransform.
+     */
+    /*@Test
```
jsmith2: Delete or ignore these tests.

Product/Production/Services/AuditRepositoryCore/src/test/java/gov/hhs/fha/nhinc/audit/transform/AuditTransformTest.java
```diff
@@ -0,0 +1,224 @@
...
+     }*/
+
+    /**
+     * Test of createEventIdentification method, of class AuditTransform.
+     *
+     * @param request
+     * @param serviceName
+     * @param isRequesting
+     */
+    public void testGetEventIdentificationType(LogEventRequestType request, String serviceName, Boolean isRequesting) {
```
jsmith2: I think all the methods in this class should probably be protected.  I think they are only meant to be accessed by sub classes.

Product/Production/Services/PatientDiscoveryCore/src/main/java/gov/hhs/fha/nhinc/patientdiscovery/audit/transform/PatientDiscoveryTransforms.java
```diff
@@ -0,0 +1,187 @@
...
+        } else {
+            LOG.error("PatientId doesn't exists in the received PRPAIN201305UV02 message");
+            return oII;
+        }
+        return oII;
+    }
+
+    private List<II> getPatientIdFromResponse(PRPAIN201306UV02 msg) {
+        List<II> oIIs = null;
+        PRPAIN201306UV02 oPatientDiscoveryResponseMessage = msg;
```
jsmith2: not sure why this line is needed, why are you assigning the variable passed in to a new variable?

Product/Production/Services/PatientDiscoveryCore/src/main/java/gov/hhs/fha/nhinc/patientdiscovery/audit/transform/PatientDiscoveryTransforms.java
```diff
@@ -0,0 +1,187 @@
...
+            marshaller.marshal(request.getControlActProcess().getQueryByParameter(), baOutStrm);
+        } else if (response != null && response.getControlActProcess() != null && response.getControlActProcess().getQueryByParameter() != null) {
+            marshaller.marshal(response.getControlActProcess().getQueryByParameter(), baOutStrm);
+        }
+
+        return baOutStrm.toByteArray();
+    }
+
+    private II getPatientIdFromRequest(PRPAIN201305UV02 msg) {
+        PRPAIN201305UV02 oPatientDiscoveryRequestMessage = msg;
```
jsmith2: Same here, new variable for argument value.

Product/Production/Services/PatientDiscoveryCore/src/main/java/gov/hhs/fha/nhinc/patientdiscovery/audit/transform/PatientDiscoveryTransforms.java
```diff
@@ -0,0 +1,187 @@
...
+import org.hl7.v3.PRPAIN201306UV02;
+import org.hl7.v3.PRPAIN201306UV02MFMIMT700711UV01Subject1;
+
+/**
+ *
+ * @author achidamb
+ * @param <T>
+ * @param <K>
+ */
+public class PatientDiscoveryTransforms<T extends PRPAIN201305UV02, K extends PRPAIN201306UV02> extends AuditTransform<T, K> {
```
jsmith2: I think you should just do PRPAIN201305UV02, PRPAIN201306UV02 and the extends is unnecessary.

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/transform/AuditTransform.java
```diff
@@ -0,0 +1,608 @@
...
+
+    /**
+     * ParticipantObjectIdentifiaction is specific to each service and the implementation can be carried out in 
+     * specific service core transformer.
+     * @param msg
+     * @param assertion
+     * @param auditMsg
+     * @return
+     */
+    protected abstract AuditMessageType getParticipantObjectIdentification(Object msg, AssertionType assertion,
```
jsmith2: You're not using the generics anymore.  Should be T msg not Object msg

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/transform/AuditTransform.java
```diff
@@ -0,0 +1,608 @@
...
+
+        // If specified, set the Audit Enterprise Site Id (community name)
+        if (communityName != null) {
+            auditSrcId.setAuditEnterpriseSiteID(communityName);
+        }
+
+        return auditSrcId;
+    }
+    
+    private LogEventRequestType createAuditMessage(Object msg, AssertionType assertion,
```
jsmith2: This needs to be refactored.  Need to either use generics or objects.  Not both.

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/transform/AuditTransform.java
```diff
@@ -0,0 +1,608 @@
...
+            XMLGregorianCalendar calendar = factory.newXMLGregorianCalendar(
+                today.get(java.util.GregorianCalendar.YEAR), today.get(java.util.GregorianCalendar.MONTH) + 1,
+                today.get(java.util.GregorianCalendar.DAY_OF_MONTH),
+                today.get(java.util.GregorianCalendar.HOUR_OF_DAY), today.get(java.util.GregorianCalendar.MINUTE),
+                today.get(java.util.GregorianCalendar.SECOND), today.get(java.util.GregorianCalendar.MILLISECOND),
+                0);
+            eventIdentification.setEventDateTime(calendar);
+        } catch (DatatypeConfigurationException e) {
+            LOG.error("DatatypeConfigurationException when creating XMLGregorian Date");
+            LOG.error(" message: " + e.getMessage(), e);
```
christophermay07: Use .getLocalizedMessage() here, and in the catch below.

Product/Production/Services/PatientDiscoveryCore/src/main/java/gov/hhs/fha/nhinc/patientdiscovery/audit/transform/PatientDiscoveryTransforms.java
```diff
@@ -0,0 +1,220 @@
...
+        Marshaller marshaller = jc.createMarshaller();
+        ByteArrayOutputStream baOutStrm = new ByteArrayOutputStream();
+        baOutStrm.reset();
+        if (msg instanceof PRPAIN201305UV02) {
+            request = (PRPAIN201305UV02) msg;
+            if (request.getControlActProcess() != null && request.getControlActProcess().
+                getQueryByParameter() != null) {
+                marshaller.marshal(request.getControlActProcess().getQueryByParameter(), baOutStrm);
+            }
+        } else if (msg instanceof PRPAIN201305UV02) {
```
christophermay07: This "else" is the same condition as the original "if" -- not sure what it's supposed to be, but take a look and fix this.

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/AuditTransformConstants.java
```diff
@@ -0,0 +1,45 @@
...
+/**
+ *
+ * @author achidamb
+ */
+public class AuditTransformConstants {
+    public static final Integer EVENT_OUTCOME_INDICATOR_SUCCESS = 0;
+    public static final String ACTIVE_PARTICPANT_USER_ID_SOURCE = "anonymous";
+    public static final String ACTIVE_PARTICPANT_ROLE_CODE_SOURCE_DISPLAY_NAME = "Source";
+    public static final String ACTIVE_PARTICPANT_ROLE_CODE_DESTINATION_DISPLAY_NAME = "Destination";
+    public static final String ACTIVE_PARTICPANT_UNKNOWN_IP_ADDRESS = "unknown";
```
christophermay07: Typo on the last four constants, should be "PARTICIPANT"

Product/Production/Gateway/PatientDiscovery_10/src/test/java/gov/hhs/fha/nhinc/patientdiscovery/_10/gateway/ws/PatientDiscoverySpringContextTest.java
```diff
@@ -89,44 +92,45 @@
...
-            faultThrown = true;
-            assertEquals("920", fault.getFaultInfo().getErrorCode());
-        }
+     inboundPDEndpoint.setInboundPatientDiscovery(inboundPatientDiscovery);
 
-        assertTrue(faultThrown);
-    }*/
+     boolean faultThrown = false;
+     try {
+     inboundPDEndpoint.respondingGatewayPRPAIN201305UV02(request);
```
christophermay07: Indentation is off here.

Product/Production/Services/AuditRepositoryCore/src/test/java/gov/hhs/fha/nhinc/audit/transform/AuditTransformTest.java
```diff
@@ -0,0 +1,213 @@
...
+import static org.junit.Assert.*;
+import org.junit.Before;
+import org.junit.BeforeClass;
+
+
+/**
+ *
+ * @author achidamb
+ */
+public abstract class AuditTransformTest<T,K> {
```
jsmith2: I don't think you use the generics in this class.  I think the generic types can be removed.

Product/Production/Services/PatientDiscoveryCore/src/main/java/gov/hhs/fha/nhinc/patientdiscovery/audit/PatientDiscoveryAuditLogger.java
```diff
@@ -0,0 +1,92 @@
...
+
+    private void auditLogMessages(LogEventRequestType auditLogMsg, AssertionType assertion) {
+        if (auditLogMsg != null && auditLogMsg.getAuditMessage() != null) {
+            audit(auditLogMsg, assertion);
+        } else {
+            LOG.error("PatientDiscoveryAuditLogger auditLogMsg is null");
+        }
+    }
+
+    /**
```
jsmith2, line 101: It's a little strange that this class has java doc for a private method and nothing for the 2 public methods.  Nitpicking.

Product/Production/Services/PatientDiscoveryCore/src/main/java/gov/hhs/fha/nhinc/patientdiscovery/audit/transform/PatientDiscoveryTransforms.java
jsmith2: there's a lot of unnecessary type declarations.  can you see the pr I sent you for how I did it?

Product/Production/Services/PatientDiscoveryCore/src/main/java/gov/hhs/fha/nhinc/patientdiscovery/audit/transform/PatientDiscoveryTransforms.java
```diff
@@ -0,0 +1,247 @@
...
+        return auditMsg;
+    }
+
+    private AuditMessageType getPatientParticipantObjectIdentificationForRequest(PRPAIN201305UV02 request, AuditMessageType auditMsg) {
+        // Set the Partipation Object Id (patient id)
+        II oII = getPatientIdFromRequest(request);
+        if (oII != null && oII.getRoot() != null && oII.getExtension() != null && !oII.getRoot().isEmpty()
+            && !oII.getExtension().isEmpty()) {
+            createPatientParticipantObjectIdentification(auditMsg, oII.getRoot(), oII.getExtension());
+        }
```
jsmith2, line 95: The response method for this has an else and with (auditMsg, null, null), should we do that here or should it be removed from the response method?

Product/Production/Services/PatientDiscoveryCore/src/main/java/gov/hhs/fha/nhinc/patientdiscovery/audit/transform/PatientDiscoveryTransforms.java
```diff
@@ -0,0 +1,247 @@
...
+            && request.getControlActProcess().getQueryByParameter().getValue().getParameterList().
+            getLivingSubjectId().get(0).getValue() != null
+            && request.getControlActProcess().getQueryByParameter().getValue().getParameterList().
+            getLivingSubjectId().get(0).getValue().get(0) != null) {
+            oII = request.getControlActProcess().getQueryByParameter().getValue().getParameterList().
+                getLivingSubjectId().get(0).getValue().get(0);
+
+        } else {
+            LOG.error("PatientId doesn't exists in the received PRPAIN201305UV02 message");
+            return oII;
```
jsmith2: remove the return statement in the else.

Product/Production/Services/PatientDiscoveryCore/src/main/java/gov/hhs/fha/nhinc/patientdiscovery/audit/transform/PatientDiscoveryTransforms.java
```diff
@@ -0,0 +1,247 @@
...
+        } else {
+            auditMsg.getParticipantObjectIdentification().add(participantObject);
+        }
+        return auditMsg;
+    }
+
+    private static String createPatientId(String assigningAuthId, String patientId) {
+
+        return patientId + "^^^&" + assigningAuthId + "&ISO";
+
```
jsmith2, line 175: extra carriage return here.

Product/Production/Services/PatientDiscoveryCore/src/main/java/gov/hhs/fha/nhinc/patientdiscovery/audit/transform/PatientDiscoveryTransforms.java
```diff
@@ -0,0 +1,247 @@
...
+    private AuditMessageType getQueryParamsParticipantObjectIdentificationForRequest(PRPAIN201305UV02 request, AuditMessageType auditMsg)
+        throws JAXBException {
+        ParticipantObjectIdentificationType participantObject = buildBaseParticipantObjectIdentificationType();
+        byte[] byteArray = getParticipantObjectQueryForRequest(request);
+        if (byteArray != null) {
+            participantObject.setParticipantObjectQuery(byteArray);
+        }
+        auditMsg.getParticipantObjectIdentification().add(participantObject);
+        return auditMsg;
+
```
jsmith2: extra carriage return here

Product/Production/Services/PatientDiscoveryCore/src/main/java/gov/hhs/fha/nhinc/patientdiscovery/audit/transform/PatientDiscoveryTransforms.java
```diff
@@ -0,0 +1,247 @@
...
+    private AuditMessageType getQueryParamsParticipantObjectIdentificationForResponse(PRPAIN201306UV02 response, AuditMessageType auditMsg)
+        throws JAXBException {
+        ParticipantObjectIdentificationType participantObject = buildBaseParticipantObjectIdentificationType();
+        byte[] byteArray = getParticipantObjectQueryForResponse(response);
+        if (byteArray != null) {
+            participantObject.setParticipantObjectQuery(byteArray);
+        }
+        auditMsg.getParticipantObjectIdentification().add(participantObject);
+        return auditMsg;
+
```
jsmith2: here too.

PR #1118, opened 2015-08-20
FHAC-480A:Modified properties and  added checks to new property and disabled newly added testcases

sailajaa:
1. Corrected property file.
2. Added checks as per property file
3. Disabled newly added testcases.


alameluchidambaram:
Reviewed PR and okay with merge.

PR #1122, opened 2015-08-31
FHAC-503:Update to PD Audit Log Regression suites

sailajaa:
1. Updated PDAuditlogging.properties file with correct values to do assertion
2. Decoded ParticipantObjectQuery value and parsed as xml message to do an assertion on each element / attribute values .
3. Updated PD service (mock response) to include semantic text.


alameluchidambaram:
Downloaded Pass through and Standard Audit Logging test suites and QA was performed before merge.

PR #1130, opened 2015-09-14
FHAC-484: Updated AuditLog testcases for DS service

sailajaa:
1. Added testcases for DS AuditLogging in AuditLog-Standard and AuditLog-Passthrough Regression project.


TabassumJafri:
Some minor things:
1. Fix the typo in script to use ActiveParticipant_Dest.@UserName instead of  ActiveParticipant_Dest.@UserNamee. 
2. Remove the property ActiveParticipant_Dest_Initiator as it is not being used.

PR #1131, opened 2015-09-14
FHAC-484: AuditLogging testcases for DS service

sailajaa:
1. Added AuditLogging testcases for DS service


TabassumJafri:
Reviewed the PR and it looks good.

PR #1135, opened 2015-09-17
FHAC-539: PD audit fixes for ATNA compliance

christophermay07:
- Updated logic for pulling correct livingSubjectId
- Set participantObjectId to the correct value, instead of a random UUID

### DIFF
--- a/Product/Production/Services/PatientDiscoveryCore/src/main/java/gov/hhs/fha/nhinc/patientdiscovery/audit/transform/PatientDiscoveryAuditTransforms.java
+++ b/Product/Production/Services/PatientDiscoveryCore/src/main/java/gov/hhs/fha/nhinc/patientdiscovery/audit/transform/PatientDiscoveryAuditTransforms.java
@@ -149,6 +149,8 @@ public class PatientDiscoveryAuditTransforms extends AuditTransforms<PRPAIN20130
 
     private II getLivingSubjectIdFromAuthorOrPerformerValue(PRPAIN201305UV02 request, List<PRPAMT201306UV02LivingSubjectId> ids) {
 
+        II livingSubjectId = null;
+        
         // Get assignedDevice root
         if (request.getControlActProcess().getAuthorOrPerformer() != null
             && !request.getControlActProcess().getAuthorOrPerformer().isEmpty()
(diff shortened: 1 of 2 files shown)


Product/Production/Services/PatientDiscoveryCore/src/main/java/gov/hhs/fha/nhinc/patientdiscovery/audit/transform/PatientDiscoveryAuditTransforms.java
```diff
@@ -129,19 +130,42 @@ private II getPatientIdFromRequest(PRPAIN201305UV02 request) {
...
+            .getLivingSubjectId() != null) {
+
+            List<PRPAMT201306UV02LivingSubjectId> ids = request.getControlActProcess().getQueryByParameter().getValue()
+                .getParameterList().getLivingSubjectId();
+
+            if (ids.size() == 1) {
+                livingSubjectId = getLivingSubjectId(ids.get(0));
+            } else if (ids.size() > 1) {
+                // Get assignedDevice root
+                String root = request.getControlActProcess().getAuthorOrPerformer().get(0).getAssignedDevice()
```
jsmith2: There are no null checks for getAuthorOrPerformer and all subsequent calls. Could throw null pointer or index out of bounds.  Are these checks done elsewhere?  Definitely would not want an exception thrown in auditing and cause message processing to fail.

Product/Production/Services/PatientDiscoveryCore/src/main/java/gov/hhs/fha/nhinc/patientdiscovery/audit/transform/PatientDiscoveryAuditTransforms.java
```diff
@@ -129,19 +130,42 @@ private II getPatientIdFromRequest(PRPAIN201305UV02 request) {
...
+                    .getValue().getId().get(0).getRoot();
+
+                // Compare assignedDevice root to each livingSubjectId root
```
jsmith2: Also, might have wanted to pull out the matching of the IDs to the root into another method. Would make the code clearer and eliminate some of the cyclomatic complexity (I think it gets to 4 levels, not too bad)

christophermay07: Agreed, but I'd argue that functionality belongs in a PD helper library (for safely traversing the JAXB object, with null checks and all), which would be out of the scope of this ticket.

PR #1136, opened 2015-09-17
FHAC-493 Updates to regression suite to fix assertions, base64 decodi…

sovanncgi:
FHAC-493 Updates to regression suite to fix assertions, base64 decoding and JMX
1. Updates to Doc Query properties files for assertion values
2. Updates to assertion script to decode base64-encoded values in audit message
3. Updates in passthrough suite to remove all JMX references


sailajaa:
1. There are no assertions on <ParticipantObjectQuery> decoded values. Atlest one assertion to compare PatientID is required.
2. Remove unused commented script lines and log.info messages in both AuditLogging-Standard and AuditLogging-Passthrough projects 
3. Why the ParticipantObjectDetail-Value and type attribute are decoded. Assertions can directly happen on ParticipantObjectDetail-Value and type attributes . This is how the DQ logging entry for ParticipantObjectDetail and they are not Base 64 encode value, so it is not required to decode in script. 
   "ParticipantObjectDetail type="QueryEncoding" value="VVRGLTg="/"
   "ParticipantObjectDetail type="urn:ihe:iti:xca:2010:homeCommunityId" value="MS4x"/"
4. Script / loop flow is wrong for ParticipantObjectDetail element.  There is an outer loop for each ParticipantObjectDetail element, and again doing xpath to specific element which is not required. Need to correct script for ParticipantObjectDetail entry: 
   
   ## "parsedXmlMessage.ParticipantObjectIdentification[1].ParticipantObjectDetail[0].each{"
   
   ## "def encodedData = parsedXmlMessage.ParticipantObjectIdentification[1].ParticipantObjectDetail[0].@value.text()"
   
   }"
5. Same comments applicable for AuditLogging-Standard -g0,g1 and AuditLogging-Passthrough-g0,g1.

christophermay07:
@sailajaa Regarding comment 3, are you saying that you're directly asserting against the values "VVRGLTg=" and "MS4x"?

christophermay07:
@sailajaa Regarding comment 1, can you be more specific about what decoded value you're referring to?

christophermay07:
@sailajaa Regarding comment 4, is there an example in another test script that can be referenced for this?

christophermay07:
@sovanncgi Regarding comment 2, any commented lines that you're responsible for either need to be commented back in (if they work), removed (if they don't ever belong), or ANOTHER comment needs to be added explaining why the comment exists and is not being removed (e.g., "// this code is will be needed when....")

If there are commented lines that you didn't add / edit, this is out of scope for the current ticket -- leave those in.

sovanncgi:
Regarding #2, the line I commented out was the assertion for alternativeuserid. Commented this out because the asserted value depends on the system initiating the message and right now asserted values are hard-coded into the properties file. In the future, we should assert for this value after we figure out how to retrieve it dynamically. For this reason, i want to keep this assertion in the test, but commented out for now. Adding comments in the code itself as to why I'm keeping this line commented out.

sailajaa:
@christophermay07 : Comment#1 : Talking about ParticipantObjectIdentification -> ParticipantObjectQuery element.

sailajaa:
@christophermay07 : Comment#3: Yes. I am talking about values "VVRGLTg=" and "MS4x". As per sec this is how they should be populated:
"In one element, set “QueryEncoding”as the value of the attribute type,Set the attribute value to the character encoding, such as “UTF-8”, used to encode the ParticipantObjectQuery before base64 encoding".
"In another element, set “urn:ihe:iti:xca:2010:homeCommunityId” as the
value of the attribute type and the value of the homeCommunityID as the value of the attribute value, if known."

My understanding, while doing assertions it is not required to decode the value.

sailajaa:
@christophermay07 : Regarding comment 4: I do not have exact example to refer, but in all scripts ( PD / RD/ QD) we are making similar checks for different elements. For Ex: EventIdentification. 

Its not going to be same, case by case the script requires to modify slightly.

christophermay07:
@sailajaa I disagree that we should compare encoded values, when it's the decoded values that are of importance.  This is a bad practice in general, especially as the encoding could be different in the future.

christophermay07:
@sailajaa I understand that the checks for QD will be different than for other services, but it seems that you copy-pasted that code snippet from somewhere.  Could you please let me know where you copied it from?

sailajaa:
@christophermay07 : As per spec, Goping by spec, my understanding is, we are going to populate values as is, not the encoded. Meaning. Homecommunictyid =1.1 not encode value, similarly for other PatiriciaptnObjectDetail entry. If this is populated as decoded, should be confirmed with developer and make checks appropriately.

sovanncgi:
Regarding #3, when the task was handed off there was already code to attempt to decode the base64 values of those elements and assert against decoded values, not encoded values. However, the code was not working so I modified it to work and now the assertion passes. @sailajaa maybe this was discussed prior to it being handed off to me. Please lverify and et me know if this no longer needs to be decoded for assertion.

sailajaa:
@christophermay07 : Nope, I have not copy pasted that part of code, as it is specific to DQ.  When I started working on AuditLog regression scripts, I prepared a base framework and made changes as per service. The base framework I did was way back when I did for X12. So, may be you can refer X12 scripts, but just an FYI, X12 done way back.

sailajaa:
@sovanncgi: I understand , the initial scripts are created to test against sample data. But, there were good amount of discussions happened as part of QD development, and we got clear idea about what actual values to populate for some of the elements / attributes.  

It is understandable that the script need  changes to make sure all the assertions are correct after the final code / design approach.  You are handling the QA task for DQ, it was never done by other person.

If i remember correct, in the original script base64 decode happened only for ParticipantObjectQuery element, not for ParticipantObjectDetail elements. 

As part of QA we have to verify spec wiki, actual message and Regression script to make sure all are in sync when the QD service is finally done.  It was never discussed before, I got your PR code into my local and testing against DQ specs, thats when I had question about ParticipantObjectDetail elements decode. When you are doing QA, you have to reach out to developer and cross verify with wiki to make sure code is logging correct message. So that you can make appropriate assertions.

sailajaa:
Reviewed PR and good with PR changes.

PR #1137, opened 2015-09-18
FHAC-517:Refactor X12 to adopt Template Design pattern like other services

alameluchidambaram:
X12 is refactored to use abstract AuditTransforms class to use functionality common across all other services. 
AuditLogging modified once for the Request in the initiator and once for the Response in the Responder for both X12 RealTime, BatchRequest and BatchResponse services.

NetworkAccessPointTypeId is modified to "2" for Destination ActiveParticipant. since it is read from target url from UDDI and it is localhost and previously it was populating "1". In AuditTransforms which is used across services is populating NetworkAccessPointTypeId "2".


Product/Production/Services/CORE_X12DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/corex12/docsubmission/audit/CORE_X12AuditDataTransformConstants.java
```diff
@@ -56,5 +53,6 @@
     public static final String CAQH_X12_CONNECTIVITY_CODED_SYS_NAME = "CAQH CORE Connectivity Metadata";
     public static final String CAQH_X12_CONNECTIVITY_CODED_SYS_DISPLAY_NAME = "Payload Identifier";
     public static final Short PARTICIPANT_OBJ_TYPE_CODE_ROLE_X12 = 20;
```
christophermay07: You used the wrapper type here, but the native type above (Short vs. short) - is there a reason for not sticking to one or the other?

Product/Production/Services/CORE_X12DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/corex12/docsubmission/audit/transform/COREX12AuditTransforms.java
```diff
@@ -0,0 +1,223 @@
...
+     * @param auditMsg
+     * @return
+     */
+    @Override
+    protected final AuditMessageType getParticipantObjectIdentificationForRequest(T request,
+        AssertionType assertion, AuditMessageType auditMsg) {
+
+        ParticipantObjectIdentificationType participantObject
+            = buildBaseParticipantObjectIdentificationType(getPayloadFromRequest(request));
+        byte[] baos = marshallToByteArrayFromRequest(request);
```
christophermay07: You don't need to change it for this PR, but for the future, baos is typically used as a name for a ByteArrayOutputStream, not a byte[]

Product/Production/Services/CORE_X12DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/corex12/docsubmission/audit/transform/COREX12AuditTransforms.java
```diff
@@ -0,0 +1,223 @@
...
+     *
+     * @return
+     */
+    protected Marshaller getMarshaller() {
+        Marshaller marshaller = null;
+        try {
+            JAXBContextHandler oHandler = new JAXBContextHandler();
+            JAXBContext jc = oHandler.getJAXBContext(CORE_X12AuditDataTransformConstants.CORE_X12_JAXB_CONTEXT);
+            marshaller = jc.createMarshaller();
+            return marshaller;
```
christophermay07: This return is redundant and should be removed

Product/Production/Services/CORE_X12DocumentSubmissionCore/src/main/java/gov/hhs/fha/nhinc/corex12/docsubmission/audit/transform/COREX12AuditTransforms.java
```diff
@@ -0,0 +1,223 @@
...
+
+    /**
+     *
+     * @param nameSpaceURI
+     * @param localPart
+     * @return
+     */
+    protected QName getQname(String nameSpaceURI, String localPart) {
+        QName xmlQname = null;
+        xmlQname = new QName(nameSpaceURI, localPart);
```
christophermay07: This function should be a single line (inline return).

jsmith2:
Approve pull request with small updates suggested by Chris.  Will merge after those updates made.

PR #1142, opened 2015-09-24
FHAC-569: RegressionScript fixes for UserID and Script clean-up

sailajaa:
1. Fixed assertions for Source-UserID (PD,DQ,X12,DS) and Destination-UserID (RD).
2. Fixed endpoints for DS, RD services for g0,g1
3. As part of clean-up removed log info statements, checks that are not required, properties that are not required.


vdmehta06:
I have tested this ticket with Standard and passthru mode for g0 and g1 suites using JBOSS app server for PD, DS, DQ , X12 and RD services and all pass successfully and good with merge

PR #1143, opened 2015-09-28
FHAC-563 Fix for RD service NetworkAccessPointId to populate host add…

vdmehta06:
userID and NetworkAccessPointID for ActiveParticipant's source and destination for initiating gateway and responding gateway are correctly populated as a part of this ticket. 
For ActiveParticipant's source NetworkAccessPointID in initiating and responding gateway  - it will populate responding gateway's IPAddress and for ActiveParticipant's destination NetworkAccessPointID - it will always populate initiating gateway's IPAddress.
For ActiveParticipant source's userID in initiating and responding gateway - it will populate the remoteURL for responding gateway 
For ActiveParticipant destination userID in initiating and responding gateway - it will populate whatever is present in wsa:ReplyTo/ SOAP header
Testing: I have done gateway to gateway testing ( from your local machine to GFE LDAP machine) and for both initiating and responding side I can see correct IPAddress populated for NetworkAccessPointID and for userID remoteURL populated. It does not require to change Junit testcase as its already testing those conditions earlier

### DIFF
--- a/Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/auditrepository/nhinc/AuditRepositoryOrchImpl.java
+++ b/Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/auditrepository/nhinc/AuditRepositoryOrchImpl.java
@@ -116,7 +116,9 @@ public class AuditRepositoryOrchImpl {
             if (activeParticipant != null) {
                 eventUserId = activeParticipant.getUserID();
                 if (eventUserId != null && !eventUserId.equals("")) {
-                    auditRec.setUserId(eventUserId);
+                    /* this value is temporary set to empty strings and due to length of characters are exceeding
+                     more than 100*/
+                    auditRec.setUserId("");
                 } else {
                     auditRec.setUserId("");
                 }
--- a/Product/Production/Services/DocumentRetrieveCore/src/main/java/gov/hhs/fha/nhinc/docretrieve/audit/transform/DocRetrieveAuditTransforms.java
+++ b/Product/Production/Services/DocumentRetrieveCore/src/main/java/gov/hhs/fha/nhinc/docretrieve/audit/transform/DocRetrieveAuditTransforms.java
@@ -333,7 +333,13 @@ public class DocRetrieveAuditTransforms
          responder side. And for this service, source is considered who generates the document, meaning that
          destination that generates the document is considered source for RD and vice versa for destination
          details for ActiveParticipant */
-        String hostDetails = getWebServiceUrlFromRemoteObject(target, serviceName);
+        String hostDetails = null;
+        if (isRequesting) {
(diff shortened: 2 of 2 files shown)


Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/auditrepository/nhinc/AuditRepositoryOrchImpl.java
```diff
@@ -116,7 +116,9 @@ public AcknowledgementType logAudit(LogEventSecureRequestType mess, AssertionTyp
             if (activeParticipant != null) {
                 eventUserId = activeParticipant.getUserID();
                 if (eventUserId != null && !eventUserId.equals("")) {
-                    auditRec.setUserId(eventUserId);
+                    /* this value is temporary set to empty strings and due to length of characters are exceeding
+                     more than 100*/
+                    auditRec.setUserId("");
```
christophermay07, line 121: Is there a limitation of 100 chars in the schema?  Is it better to set blank, than to truncate?

Product/Production/Services/DocumentRetrieveCore/src/main/java/gov/hhs/fha/nhinc/docretrieve/audit/transform/DocRetrieveAuditTransforms.java
```diff
@@ -333,7 +333,13 @@ protected ActiveParticipant getActiveParticipantSource(NhinTargetSystemType targ
          responder side. And for this service, source is considered who generates the document, meaning that
          destination that generates the document is considered source for RD and vice versa for destination
          details for ActiveParticipant */
-        String hostDetails = getWebServiceUrlFromRemoteObject(target, serviceName);
+        String hostDetails = null;
```
christophermay07, line 336: Don't create another commit for this, but in the future, don't initialize a variable (even if it's just to null) if you're guaranteed to set the value before you try to dereference.

christophermay07:
I'm good with this PR, pending the answer to one question.

Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/auditrepository/nhinc/AuditRepositoryOrchImpl.java
alameluchidambaram, line 121: We have to re-design audit database and the userName is defined to 100 characters which is enough if we populate the HumanRequestor userId. Currently it always populates the first ActivaParticipant userId for both initiating and Responding Gateway and in this case the service endpoint is the source ActivePartcicipant UserId. This will be addressed as part of FHAC-511 which is database design for audit db.

alameluchidambaram:
Good with PR since the junits are going to be fixed as part of next pr.

PR #1146, opened 2015-09-30
FHAC-547:xml-api and xercesImpl jars are modified to provided scope to use jdk provided jars during runtime

alameluchidambaram:
1. xml-api and xercesImpl jars modified from compile, runtime to scope provided.
2. Tested on weblogic 12.1.3,websphere 8.5.5.3, wildfly 8.2.1-Final on Windows

### DIFF
--- a/Product/pom.xml
+++ b/Product/pom.xml
@@ -577,6 +577,10 @@
                         <groupId>xerces</groupId> 
                         <artifactId>xercesImpl</artifactId> 
                     </exclusion>
+                    <exclusion>
+                        <groupId>xml-apis</groupId>
+                        <artifactId>xml-apis</artifactId>
+                    </exclusion>
                 </exclusions>
             </dependency>
             <dependency>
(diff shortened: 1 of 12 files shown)


Product/pom.xml
```diff
@@ -292,6 +292,7 @@
                 <groupId>xerces</groupId>
                 <artifactId>xercesImpl</artifactId>
                 <version>2.10.0</version>
+                <scope>provided</scope>
```
christophermay07: Scope of "provided" is not the right way to do this, we should exclude the jars where they are transitive dependencies.

jsmith2:
Looks good to me.  As long as this has been confirmed on all app servers, I approve with no comments.

christophermay07:
`org.connectopensource:CONNECTAdminGUI:war:4.6.0-SNAPSHOT`
`\- me.fhir:fhir-0.12:jar:0.3:compile`
`\- net.sf.saxon:Saxon-HE:jar:9.4:compile`
`\- xom:xom:jar:1.2.5:compile`
`\- xerces:xercesImpl:jar:2.8.0:compile`
Is this going to cause a problem if AdminGUI is deployed alongside CONNECT on WLS?

sovanncgi:
This has been verified on all app servers.
https://cgiinterop.atlassian.net/browse/FHAC-587 has been created to address potential AdminGUI issues

PR #1149, opened 2015-10-01
FHAC-578: removed the ParticipantObjectName assertion from AuditLoggi…

TabassumJafri:
1. Removed the ParticipantObjectName assertion from AuditLogging Standard and Passthru Regression Suite.
2. ParticipantObjectName is not removed from DR AuditLogging steps (in both Standard and Passthru)that are marked as ToBeRemoved.


sailajaa:
Reviewed changes and good with PR merge.

PR #1151, opened 2015-10-03
FHAC-546:Fix for Glassfish Application Server Audit Logging

alameluchidambaram:
1. Ear packaging changed into Application server specific pom since Glassfish requires EJB jar to be bundled in lib directory.


christophermay07:
I'm good with this PR as I don't see another way to get this working on GlassFish, however -- when GlassFish is no longer supported by CONNECT, we should consider reverting this change (this is non-standard behavior, and undoes the whole idea of using the individual app server POMs only to include/exclude specific dependencies).

sailajaa:
I reviewed and tested PR locally on GlassFish. Good with PR merge.

PR #1152, opened 2015-10-05
FHAC-591:Remove xml-apis dependency from Direct module

alameluchidambaram:
Remove xml-apis jar from Direct module. 
CONNECT ear is built without xml-apis and xercesImpl jar


christophermay07:
Approved, but we need to make sure this ticket is considered part of last sprint, since it's going to be under the last sprint tag

sailajaa:
Reviewed and good with PR changes.

PR #1154, opened 2015-10-08
FHAC-582: Regression Script changes for DR service

sailajaa:
1. Updated script for DR service to test two entries, one for initiator and one for responder.
2. While testing found issue with ParticipantObjectDetail - value element. created ticket and commented that assertions.
3. Updated script for PD service to increase delay time and to add NetworkAccessPointID assertions.


vdmehta06:
I have tested this PR with Passthrough and Standard mode for PD and DR. Good with merge after removing GatewayPropDir value which is hard coded in SOAPUI xml for both Standard and Passthrough mode.

vdmehta06:
Good with change and going to merge.

PR #1157, opened 2015-10-14
FHAC-538 AuditLogging Regression Suite tests for PD Deferred Service

vdmehta06:
Regression test suites are created for Standard and Passthrough for PD Deferred Service. It is tested with sample SQL blob message. Basically, for PD Deferred - test suite will verify two entries which will be generated for PD Deferred Request ( one entry on initiating end and other one on responding gateway ) and verify two entries which will be generated for PD Deferred Response ( one entry on initiating end and other one on responding gateway ).


sailajaa:
Disable VerifyMessage test steps as we are currently testing with sample data and enable PD deferred service test step.

sailajaa:
Add Delay test step for Passthrough

sailajaa:
Reviewed and good with PR merge.

PR #1158, opened 2015-10-14
FHAC-597: Regression Script updated for DS service

sailajaa:
1. Regression Script updated for DS service in AuditLogging-Standard, AuditLogging-Passthrough Projects.


TabassumJafri:
Not sure why the ParticipantObjectID assertion is commented out for Submission Set ParticipantObjectIdentification inboth DS Audit Logging Standard and Passthru,

sailajaa:
Submission Set ParticipantObjectID asertion is commented out, because when gateway-gateway test happens this value would be dynamic and assertion may fail. We thought of doing a manual check for this element comparison.

TabassumJafri:
PR looks good and ready to merge.

PR #1163, opened 2015-10-23
FHAC-616 Remove Fortify "Dead Code" Unused Methods and Fields

vdmehta06:
Addressed Fortify findings regarding unused methods and fields

### DIFF
--- a/Product/Production/Adapters/General/CONNECTAdapterPDPOpenSSO/src/main/java/com/sun/identity/xacml/plugins/XSPAXACMLAuthzDecisionQueryHandler.java
+++ b/Product/Production/Adapters/General/CONNECTAdapterPDPOpenSSO/src/main/java/com/sun/identity/xacml/plugins/XSPAXACMLAuthzDecisionQueryHandler.java
@@ -74,7 +74,6 @@ import java.util.Date;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 
 import org.w3c.dom.Document;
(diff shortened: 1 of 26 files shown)


Product/Production/Services/PolicyEngineCore/src/main/java/gov/hhs/fha/nhinc/policyengine/adapter/pip/XACMLCreator.java
```diff
@@ -360,11 +329,10 @@ else if ((sHL7DateTime != null) && (sHL7DateTime.length() > 8)) {
                 sXMLDate = oXMLDateTimeFormatter.format(oHL7Date);
             } catch (Exception e) {
                 String sErrorMessage = "Date-time had invalid HL7 format: " + sHL7DateTime + ".  Error = "
-                        + e.getMessage();
+                    + e.getMessage();
                 throw new AdapterPIPException(sErrorMessage, e);
             }
-        }
-        // Do not know format - put it in as is.
+        } // Do not know format - put it in as is.
```
christophermay07, line 335: Making the following comment as an FYI for future reference, DON'T make another commit to change this:

I know you made this change to stay consistent with the rest of the class, but it's actually the rest of the class that's wrong -- comments should always be on their own lines.  In this case, I would have suggested moving "// Do not know format..." into the else, right about the "sXMLDate" assignment.  That way, the comment gets its own line, it's unambiguous where and why the comment belongs, and as an afterthought (but still an added bonus) -- it helps our CheckStyle finding count :)

christophermay07:

Product/Production/Adapters/General/CONNECTAdapterPDPOpenSSO/src/main/java/com/sun/identity/xacml/plugins/XSPAXACMLAuthzDecisionQueryHandler.java
```diff
@@ -286,19 +284,18 @@ public XSPAXACMLAuthzDecisionQueryHandler() {
         }
 
         com.sun.identity.saml2.protocol.Response samlpResponse = createSamlpResponse(statement, status.getStatusCode()
-                .getValue());
+            .getValue());
 
         System.out.println("XSPAXACMLAuthzDecisionQueryHandler.handleQuery(), returning " + ":samlResponse=\n"
-                + samlpResponse.toXMLString(true, true));
+            + samlpResponse.toXMLString(true, true));
```
TabassumJafri: Since we are already attaching this file, is it possible to remove System.out.println lines and replace them with log statements

TabassumJafri, line 290: Since we are already touching this file, is it possible to remove System.out.println lines and replace them with log statements

christophermay07, line 290: @TabassumJafri Great catch, and I agree!! But, this is a special case -- "System.out/exit" calls will be taken out of as part of cleanup tasks surrounding SonarQube and CheckStyle during an upcoming sprint -- so, no need to add another commit for it now.  In the meantime, keep looking out for little things we can do to make the code better :)

PR #1164, opened 2015-10-24
FHAC-499 Fix UDDI and ICI xml files for Direct testing

sovanncgi:
Automated Direct has been failing due to a regression test replacing the internalconnectioninfo.xml file with one that does not have Direct endpoints. To address this, the Automated Direct test suite has been updated to make sure the standard internalconnectioninfo.xml and the UDDI files are used for Direct testing:
1. Added internalConnectionInfo_g1.xml
2. Added uddiConnectionInfo_g1.xml
3. Made changes in SoapUI script to replace existing files with these new files
4. Changed the name of the Automated Direct test to remove the word "outbound" since the test asserts for both outbound and inbound conditions


sailajaa:
I do not see 4th comment changes in PR. When I get the PR code locally, I see project name as "AutomatedDirectOutboundTesting" .

sailajaa:
There is a .properties file under Direct test project, need to rename .properties file for naming consistency purpose.

sailajaa:
There is a pom.xml under AutomatedDirectOutboundTesting. Module name in pom.xml also need to be updated for this project.

vdmehta06:
I don't see any internalConnectionInfo_g0.xml and uddiConnectionInfo_g0.xml files under the project directory. So, can we have name without specification indication like uddiConnectionInfo.xml and uddiConnectionInfo.xml instead?

sovanncgi:
Sailaja, all PR comments have been addressed.

sovanncgi:
Vidhi, understand what you're saying but I want to make it clear it's the same g1 ici and uddi files as what's in the validation suite. Going to keep the g1 extensions for this reason.

vdmehta06:
I am good with merge and have no further comments

PR #1170, opened 2015-10-30
FHAC-581:Auditlogging Regression Suite fix to remove check for order of entries logged in DB

sailajaa:
1. Removed check to verify other of entires logged in database in AuditLogging-Standard, AuditLogging-Passthrough RegressionSuites.
   2.Reduced delay time.


TabassumJafri:
Reviewed PR and good with merge

sovanncgi:
Reviewed PR. Assertions no longer dependent on record ID, therefore order of audit log events no longer matter. AD still has assertions based on order but no changes were made to AD audit logging itself so that is acceptable. Good with merge.

PR #1179, opened 2015-11-12
FHAC-568 Create AuditLog off suite for testing auditlogging is turned off

vdmehta06:
Regression suite Audit Logging On / Off changes are done for both Standard and Passthrough mode. Assertions and check is made to see whether audit logging is off and on for each service. Based on audit logging status from audit.properties check is made to verify number of database entries.
With this ticket, changes related to FHAC-654 is also done as a part of this commit. Changes are made for both Standard and Passthrough mode for replacing communityid with remoteHcid and messageType with direction field

### DIFF
--- a/Product/SoapUI_Test/RegressionSuite/Passthru/AuditLogging-Passthrough/AuditLogging-Passthrough-soapui-project.xml
+++ b/Product/SoapUI_Test/RegressionSuite/Passthru/AuditLogging-Passthrough/AuditLogging-Passthrough-soapui-project.xml
@@ -646,6 +646,7 @@ else if ( testRunner.testCase.getPropertyValue("auditLogStatus") == "false")
  	context.withSql(context.findProperty('AuditRepoDB')) {sql ->
 	assert 0 == sql.firstRow('select count(*) from ' + context.findProperty('AuditRepoTable'))[0]
 	}
+	testRunner.testCase.setPropertyValue('AuditOff', 'true')
 	testRunner.gotoStepByName( "Success")
 }
 else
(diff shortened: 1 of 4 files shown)


sailajaa:
Reviewing PR.

sailajaa:
Since It is a huge .xml file I got the PR branch into my local for testing the changes. I deployed latest CONNECT.ear and updated AuditLogging-Passthrough.xml with current PR code. When I run the First test case (Document Submission Deferred Req) , it is failing and I see " Incorrect AuditLogging Property found for this service testcase " error in SoapUI scriptLog and I see a nullpointer in SoapUI errorlog.

sailajaa:
As per my observation "auditLogStatus" property value is getting reset by SoapUI to empty value after first run. No matter if its false / true. I think ""auditLogStatus" need to be reset to it's appropriate value for all the if / if else loops that are there in "Verify audit log" step.

sailajaa:
Default value for "DocSubmissionDeferredReq" property are any other properties that are defined in audit.properties is "true". We are reading this property in setup script and resetting value to "false" . I think we need to reset back to "true" (default value) for all those audit.properties in respective testcases / test steps.

sailajaa:
I think we do not need below else condition in scripts. Because if auditlog is off, and still audittable updated with audit entries, then assertion == 0  loop fails and that itself is indication that auditlog off not working as expected. 
Here is the else loop that I think is not required in scripts:
else {
    testRunner.fail( "Incorrect AuditLogging property found for this service testcase" )
}

christophermay07:
I see these on the two SoapUI project files:

134,150 additions
90,442 additions

Did you attach the interfaces / schemas as well?  If so, please remove, commit, and push.

sailajaa:
GatewayPropDir value is present at project level. Can you pls make sure to remove that value

sailajaa:
I ran AuditLog-Passthrough project with PR code. When I run couple of times first test case is failing and not going to other testcase. Will have to check with @vdmehta06

christophermay07:
I'm seeing some minor inconsistencies between the properties in the SoapUI project vs. the props file.  Starting at the project-level custom properties, and then once for each test case, can you reload custom properties from file?

sailajaa:
Ran into few issues with current approach that handled on/off suite testcases. Suggested @vdmehta06  with changes that are required and she is going to update PR.

sailajaa:
This PR is validated by getting PR branch to a local repo. AuditLogging on / off suit passes for all services. This PR also includes changes related FHAC-652 , FHAC-654. Reviewed these changes too.

sailajaa:
I amd good with PR merge.

PR #1181, opened 2015-11-16
FHAC-194

christophermay07:
Abstracted logging to use SLF4J, with a default runtime dependency of log4j12.


christophermay07:
@vdmehta06 @alameluchidambaram Assigned for review.

vdmehta06:
I see in FileUtils.java file under SOAPUI_Test project log4j.Logger is used. Is that need to be there or need to be removed?

christophermay07:
@vdmehta06 FileUtils is used as a supporting library for our SoapUI tests, and SoapUI uses log4j natively.  For this PR, I was only concerned with code that will (or could) be deployed as part of the CONNECT gateway / reference adapters.

alameluchidambaram:
When I search for log4j in CONNECT folder I came across three instances of it

Can you modify or remove Javadoc comments?

Product\Production\Common\CONNECTCoreLib\src\main\java\gov\hhs\fha\nhinc\transform\policy\PatientDiscoveryPolicyTransformHelper.java:     \* log4j error messages are logged here
Product\Production\Common\CONNECTCoreLib\src\main\java\gov\hhs\fha\nhinc\transform\policy\PatientDiscoveryPolicyTransformHelper.java:     \* log4j debug messages are logged here
Product\SoapUI_Test\FileUtils\src\main\java\nhinc\FileUtils.java:import org.apache.log4j.Logger;

christophermay07:
@alameluchidambaram FileUtils is expect; see my earlier comment to @vdmehta06 .  As for the other two places, good catch, I was only checking code.  Those functions are horrifying, however, so I'll be removing them altogether.

alameluchidambaram:
We discuss about doing WebLogic Installation. Will hold on approving this PR until WebLogic testing is successful

alameluchidambaram:
I have tested this PR on WebLogic 12.1.3 and passes Validation suite.

vdmehta06:
While deploying Connect EAR to Weblogic 12.1.1, I found that two jars - log4j and log4j-over-slf4j exist in the lib directory and it was causing problem to deploy Connect on that server. So Chris removed in his new commit.
After his fix, I found other thing - java.lang.ClassNotFoundException: weblogic.logging.log4j.JDKLog4jAdapterFactory while starting up Weblogic server and found <StdErr> <BEA-000000> <SLF4J: Class path contains multiple SLF4J bindings.> while deploying Connect EAR. But I can run connect Validation suite without any problem. So do we need to take a look for this?

vdmehta06:
Even if I see those log4j errors, as mentioned above in the comment, Audit logging works for Standard and Passthrough mode as expected and logging two entries and passing all the services for both g0 and g1 suites and keeping that in mind and tested fully on Weblogic 12.1.1 version.
But question is why does it shows 301 files changed after your rebase instead of 612 file changes?

alameluchidambaram:
I have tested this PR on WebSphere 8.5.5.3 and looks good.

alameluchidambaram:
I have reviewed this PR and good with merge.

vdmehta06:
I have tested this PR on Weblogic 12.1.1 version and logging works properly before and after this PR. I am good with merge without any further comment

PR #1190, opened 2015-12-03
MessageId and relatesTo fix

TabassumJafri:
1. Fixed BaseService class to first check for MessageId in assertion block for the incoming request. If it is not present, next we look at SOAP headers. If messageId is not present in SOAP header then connect will create one and add it to assertion block.
2. Fixed relatesTo, to not append the incoming value with urn:uuid prefix.
3. Fixed the PD and DQ for fan-out scenario and updated the concerned junit test.

PR assignee:  @jsmith2 and @alameluchidambaram

### DIFF
--- a/Product/Production/Services/PatientDiscoveryCore/src/test/java/gov/hhs/fha/nhinc/patientdiscovery/outbound/deferred/request/StandardOutboundPatientDiscoveryDeferredRequestTest.java
+++ b/Product/Production/Services/PatientDiscoveryCore/src/test/java/gov/hhs/fha/nhinc/patientdiscovery/outbound/deferred/request/StandardOutboundPatientDiscoveryDeferredRequestTest.java
@@ -75,6 +75,7 @@ import org.junit.Test;
 import org.mockito.ArgumentCaptor;
 import static org.mockito.Matchers.eq;
 import static org.mockito.Matchers.isNull;
+import static org.mockito.Mockito.atLeast;
 import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.verify;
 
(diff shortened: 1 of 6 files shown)


Product/Production/Services/PatientDiscoveryCore/src/test/java/gov/hhs/fha/nhinc/patientdiscovery/outbound/deferred/request/StandardOutboundPatientDiscoveryDeferredRequestTest.java
```diff
@@ -271,7 +272,7 @@ public void policyFailed() throws ConnectionManagerException {
         verify(pd201305Processor).storeLocalMapping(requestArgument.capture());
         assertEquals(request, requestArgument.getValue().getPRPAIN201305UV02());
 
-        verify(correlationDao).saveOrUpdate("urn:uuid:messageId", patientId);
```
jsmith2, line 274: For an outbound message, are we still prepending the appropriate prefix ("urn:uuid") to the message Id for all services?  I think that is still a requirement.

TabassumJafri, line 274: For an Outbound request, if assertion.messageId value is present, we don't format it to have urn:uuid: prefix. The concerned test-case reflects this scenario.

alameluchidambaram:

PR #1208, opened 2015-12-30
Fhac 747

TabassumJafri:
1.  User is now able to enter date and time in both Start Time and End Time fields on the GUI.
2. After Search, when user clicks on Audit XML, only then the Audit blob is fetched and displayed on screen.
3. Fixed formatting and alignment issues on the screen.

@christophermay07, @vdmehta06 and @mhpnguyen

### DIFF
--- a/Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/services/impl/AuditServiceImpl.java
+++ b/Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/services/impl/AuditServiceImpl.java
@@ -71,18 +71,11 @@ public class AuditServiceImpl implements AuditService {
 
     private static final Logger LOG = LoggerFactory.getLogger(AuditServiceImpl.class);
     private static final String EVENT_DATE_FORMAT = "MM/dd/yyyy HH:mm:ss";
+    private static final String AUDIT_MESSAGE_NAMESPACE = "com.services.nhinc.schema.auditmessage";
     private AuditRetrieve auditRetrieve = null;
 
     public AuditServiceImpl() {
         auditRetrieve = new AuditQueryLogProxyObjectFactory().getAuditRetrieveProxy();
-    }
-
-    @Override
-    public String createMockAuditMessage(long id) {
-        Audit audit = new Audit();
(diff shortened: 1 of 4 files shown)


Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/services/impl/AuditServiceImpl.java
```diff
@@ -213,8 +207,9 @@ private String marshallAuditMessage(AuditMessageType mess) {
         if (mess != null) {
             try {
                 JAXBContextHandler oHandler = new JAXBContextHandler();
-                JAXBContext jc = oHandler.getJAXBContext("com.services.nhinc.schema.auditmessage");
+                JAXBContext jc = oHandler.getJAXBContext(AUDIT_MESSAGE_NAMESPACE);
                 Marshaller marshaller = jc.createMarshaller();
+                marshaller.setProperty(Marshaller.JAXB_FORMATTED_OUTPUT, true);
                 ByteArrayOutputStream baOutStrm = new ByteArrayOutputStream();
```
mhpnguyen: If exception happens, the ByteArrayOutputstream will never close it.  Please move it to final block.

Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/services/impl/AuditServiceImpl.java
```diff
@@ -213,8 +207,9 @@ private String marshallAuditMessage(AuditMessageType mess) {
         if (mess != null) {
             try {
                 JAXBContextHandler oHandler = new JAXBContextHandler();
-                JAXBContext jc = oHandler.getJAXBContext("com.services.nhinc.schema.auditmessage");
+                JAXBContext jc = oHandler.getJAXBContext(AUDIT_MESSAGE_NAMESPACE);
```
vdmehta06: You can do something like new JAXBContextHandler().getJAXBContext(AUDIT_MESSAGE_NAMESPACE); instead of creating JAXBContextHandler object

vdmehta06:
Good with PR changes and endorsing to merge

PR #1212, opened 2016-01-12
FHAC-791: Null dereference fixes

christophermay07:
Null dereference fixes, surrounding code cleanup

### DIFF
--- a/Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/docrepository/adapter/AdapterComponentDocRepositoryHelper.java
+++ b/Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/docrepository/adapter/AdapterComponentDocRepositoryHelper.java
@@ -38,38 +38,37 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import javax.activation.DataHandler;
+import oasis.names.tc.ebxml_regrep.xsd.rim._3.ClassificationType;
+import oasis.names.tc.ebxml_regrep.xsd.rim._3.SlotType1;
 import oasis.names.tc.ebxml_regrep.xsd.rs._3.RegistryError;
+import org.apache.commons.lang.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-
 public class AdapterComponentDocRepositoryHelper {
 
-	private static final String VALUE_LIST_SEPERATOR = "~";
-	private static final Logger LOG = LoggerFactory.getLogger(AdapterComponentDocRepositoryHelper.class);
+    private static final String VALUE_LIST_SEPERATOR = "~";
+    private static final Logger LOG = LoggerFactory.getLogger(AdapterComponentDocRepositoryHelper.class);
 
-	RegistryError setRegistryError(String codeContext, String location, String errorCode,
-    		String value){
-    	RegistryError error = new oasis.names.tc.ebxml_regrep.xsd.rs._3.ObjectFactory().createRegistryError();
-    	error.setCodeContext("ProvideAndRegisterDocumentSetRequest message handler did not " + codeContext);
-    	error.setLocation("DocumentRepositoryService.documentRepositoryProvideAndRegisterDocumentSetB -> "
-                + "DocumentRepositoryHelper.documentRepositoryProvideAndRegisterDocumentSet" + location);
-    	error.setErrorCode(errorCode);
-    	error.setSeverity(NhincConstants.XDS_REGISTRY_ERROR_SEVERITY_ERROR);
+    RegistryError setRegistryError(String codeContext, String location, String errorCode, String value) {
+        RegistryError error = new oasis.names.tc.ebxml_regrep.xsd.rs._3.ObjectFactory().createRegistryError();
+        error.setCodeContext("ProvideAndRegisterDocumentSetRequest message handler did not " + codeContext);
+        error.setLocation("DocumentRepositoryService.documentRepositoryProvideAndRegisterDocumentSetB -> "
+            + "DocumentRepositoryHelper.documentRepositoryProvideAndRegisterDocumentSet" + location);
+        error.setErrorCode(errorCode);
+        error.setSeverity(NhincConstants.XDS_REGISTRY_ERROR_SEVERITY_ERROR);
         error.setValue(value);
 
-        LOG.error("Error Location: " + error.getLocation() + "; \n" + "Error Severity: " + error.getSeverity()
-                + "; \n" + "Error ErrorCode: " + error.getErrorCode() + "; \n" + "Error CodeContext: "
-                + error.getCodeContext());
+        LOG.error("Error Location: {}; \nError Severity: {}; \nError ErrorCode: {}; \nError CodeContext: {}",
+            error.getLocation(), error.getSeverity(), error.getErrorCode(), error.getCodeContext());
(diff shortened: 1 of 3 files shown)


Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/docrepository/adapter/AdapterComponentDocRepositoryHelper.java
```diff
@@ -149,40 +148,48 @@ String extractClassificationMetadata(
...
+                    default:
+                        break;
                 }
-                break; // found desired classification, have values, exit loop
-            } // if (classificationSchemeUUID.equals(classificationSchemeName))
+                // found desired classification, have values, exit loop
+                break;
+            }
         }
         if (classificationValue != null && !classificationValue.isEmpty()) {
```
mhpnguyen: Optional: since we already have NullChecker class, should we utilize to use some method(isNotNullishIgnoreSpace) in there? or using StringUtil from apache package?

christophermay07: I've never been a fan of the NullChecker class, but StringUtils might be a nice addition in the future.  I'll probably make this change if I have to create a second commit for another reason, but I won't create a new commit for that alone.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/docrepository/adapter/AdapterComponentDocRepositoryHelper.java
```diff
@@ -192,51 +199,33 @@ String extractClassificationMetadata(
...
-                    Iterator<String> iter = slot.getValueList().getValue().iterator();
-                    while (iter.hasNext()) {
-                        String value = iter.next();
-                        slotValues.append(value);
-                        counter++;
-                        if (counter < listSize) {
-                            slotValues.append(VALUE_LIST_SEPERATOR);
-                        }
-                    }
+                LOG.debug("Found {}: {}", slotName, slot.getValueList().getValue());
```
TabassumJafri: This statement slot.getValueList().getValue()) could cause an NPE

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/docrepository/adapter/AdapterComponentDocRepositoryHelper.java
```diff
@@ -192,51 +199,33 @@ String extractClassificationMetadata(
...
+                if (valueIndex < 0) {
+                    slotValue = StringUtils.join(slot.getValueList().getValue(), VALUE_LIST_SEPERATOR);
+                } else if (slot.getValueList() != null && slot.getValueList().getValue() != null
```
mhpnguyen: Optional: since we already have NullChecker class, should we utilize to use some method(isNotNullish) in there?

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/docrepository/adapter/AdapterComponentDocRepositoryHelper.java
```diff
@@ -248,23 +237,22 @@ String extractMetadataFromSlots(List<oasis.names.tc.ebxml_regrep.xsd.rim._3.Slot
...
      */
     String extractPatientInfo(List<oasis.names.tc.ebxml_regrep.xsd.rim._3.SlotType1> documentSlots,
-            String patientInfoName) {
+        String patientInfoName) {
         String slotValue = null;
 
-        for (oasis.names.tc.ebxml_regrep.xsd.rim._3.SlotType1 slot : documentSlots) {
-            if (DocRepoConstants.XDS_SOURCE_PATIENT_INFO_SLOT.equals(slot.getName())) {
+        for (SlotType1 slot : documentSlots) {
+            if (slot.getName().equals(DocRepoConstants.XDS_SOURCE_PATIENT_INFO_SLOT)) {
```
TabassumJafri: Potential NPE.  if(slot!=null && slot.getName()!=null && slot.getName().equals(DocRepoConstants.XDS_SOURCE_PATIENT_INFO_SLOT))

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/docrepository/adapter/AdapterComponentDocRepositoryHelper.java
christophermay07: Nice catch!  Not only that, but .getValue() is a List -- going to remove that from the debug statement altogether.

christophermay07: On second examination, can that cause an NPE?  ValueList is a a required field of SlotType1, and slot is guaranteed to not be null at this point.

(Still going to remove printing the List, but would like a second opinion on if that can cause NPE).

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/docrepository/adapter/AdapterComponentDocRepositoryHelper.java
christophermay07: Agreed on checking slot for null.  As for .getName(), same question as above -- that can't be null here, can it?

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/docrepository/adapter/AdapterComponentDocRepositoryHelper.java
christophermay07: Looking at the logic here, we need a stronger case than checking for emptiness (fix coming in next commit), so NullChecker won't apply here.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/docrepository/adapter/AdapterComponentDocRepositoryHelper.java
TabassumJafri: if slot is a not null and slot.getValueList() is a required field, then we just need to remove getValue() part.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/docrepository/adapter/AdapterComponentDocRepositoryHelper.java
TabassumJafri: yes you are right. Name is a required field in oasis.names.tc.ebxml_regrep.xsd.rim._3.SlotType, so we can just check for slot!=null

TabassumJafri:

mhpnguyen:
Review PR and endorse for merge without comment

PR #1218, opened 2016-01-13
FHAC-796: Modified Destination-UserId assertions for DS services

sailajaa:


sailajaa:
@mhpnguyen : PR assigned for review

PR #1237, opened 2016-01-25
FHAC-816: fix userId in audit table for DSDefResponse and PDDefResponse

mhpnguyen:
Populate userId from either Audit message or Assertion if not found.
@alameluchidambaram @vdmehta06: Can you please review this?

### DIFF
--- a/Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/transform/AuditTransforms.java
+++ b/Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/transform/AuditTransforms.java
@@ -26,12 +26,6 @@
  */
 package gov.hhs.fha.nhinc.audit.transform;
 
-import com.services.nhinc.schema.auditmessage.AuditMessageType;
-import com.services.nhinc.schema.auditmessage.AuditMessageType.ActiveParticipant;
-import com.services.nhinc.schema.auditmessage.AuditSourceIdentificationType;
-import com.services.nhinc.schema.auditmessage.CodedValueType;
-import com.services.nhinc.schema.auditmessage.EventIdentificationType;
-import com.services.nhinc.schema.auditmessage.ParticipantObjectIdentificationType;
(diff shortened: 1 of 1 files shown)


Product/Production/Services/AuditRepositoryCore/src/main/java/gov/hhs/fha/nhinc/audit/transform/AuditTransforms.java
```diff
@@ -112,7 +116,7 @@ public final LogEventRequestType transformRequestToAuditMsg(T request, Assertion
         return buildLogEventRequestType(auditMsg, direction, getMessageCommunityId(assertion, target, isRequesting),
             serviceName, assertion, auditMsg.getEventIdentification().getEventID().getDisplayName(),
             auditMsg.getEventIdentification().getEventOutcomeIndicator(),
-            auditMsg.getEventIdentification().getEventDateTime(), getUserId(auditMsg.getActiveParticipant()));
+            auditMsg.getEventIdentification().getEventDateTime(), getRequestorUserId(auditMsg,assertion));
```
vdmehta06: Is there a reason you are passing whole auditMsg object? As you just need auditMessageType.getActiveParticipant() in this method

mhpnguyen: I guess it is easy to check on null on whole object rather than its member.  We can pass getActiveParticipant object into the method but that assume the auditMsg is not null.

vdmehta06: I think you can check before the return statement about null criteria for auditMsg, as auditMsg is used for buildLogEventRequestType method's parameters

mhpnguyen:
Update code to pass list of Active Participants instead of audit Message type.

vdmehta06:
Endorsing to merge with no further comments

alameluchidambaram:

PR #1242, opened 2016-01-31
FHAC-826: Fix Direct message monitoring on WildFly

christophermay07:
Workaround for Direct message monitoring race condition.

### DIFF
--- a/Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/DirectAdapterFactory.java
+++ b/Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/DirectAdapterFactory.java
@@ -43,12 +43,20 @@ public class DirectAdapterFactory extends DirectAdapterEntity {
     private static final String BEAN_NAME_MANAGE_TASK_SCHEDULER = "manageTaskScheduler";
 
     /**
-     * Register Handlers will invoke getInstance, thereby loading the spring
-     * context and task scheduler for polling mail servers.
+     * Register Handlers will invoke getInstance, thereby loading the spring context and task scheduler for polling mail
+     * servers.
      */
     public void registerHandlers() {
-        //initialize the HibernateUtil when the Direct Servlet is initialized.. DO NOT Remove this.
-        SessionFactory session = gov.hhs.fha.nhinc.event.persistence.HibernateUtil.getSessionFactory();
+        /**
+         * TODO: There is a race condition on accessing the SessionFactory for both Direct Message Monitoring a Event
+         * Persistence; initializing both when the Direct Servlet is initialized as a workaround.
+         * <br/>
+         * DO NOT remove either of the following two lines of code until this issue is resolved.
+         */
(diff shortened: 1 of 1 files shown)


christophermay07:
@alameluchidambaram @vdmehta06 please review.

alameluchidambaram:
I have reviewed this PR and good with merge.

Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/DirectAdapterFactory.java
```diff
@@ -43,12 +43,20 @@
...
-        //initialize the HibernateUtil when the Direct Servlet is initialized.. DO NOT Remove this.
-        SessionFactory session = gov.hhs.fha.nhinc.event.persistence.HibernateUtil.getSessionFactory();
+        /**
+         * TODO: There is a race condition on accessing the SessionFactory for both Direct Message Monitoring a Event
+         * Persistence; initializing both when the Direct Servlet is initialized as a workaround.
+         * <br/>
+         * DO NOT remove either of the following two lines of code until this issue is resolved.
+         */
+        SessionFactory eventSession = gov.hhs.fha.nhinc.event.persistence.HibernateUtil.getSessionFactory();
+        SessionFactory msgMonitorSession = gov.hhs.fha.nhinc.direct.messagemonitoring.persistence.HibernateUtil
```
vdmehta06, line 57: Why do we have full path? Can we just have import statement instead?

christophermay07, line 57: There are two "HibernateUtil" classes used back-to-back, and we can only import one (at most) anyway.  I think it would be more confusing to import just one, rather than to fully qualify both.

Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/DirectAdapterFactory.java
```diff
@@ -43,12 +43,20 @@
...
     public void registerHandlers() {
-        //initialize the HibernateUtil when the Direct Servlet is initialized.. DO NOT Remove this.
-        SessionFactory session = gov.hhs.fha.nhinc.event.persistence.HibernateUtil.getSessionFactory();
+        /**
+         * TODO: There is a race condition on accessing the SessionFactory for both Direct Message Monitoring a Event
+         * Persistence; initializing both when the Direct Servlet is initialized as a workaround.
+         * <br/>
+         * DO NOT remove either of the following two lines of code until this issue is resolved.
+         */
+        SessionFactory eventSession = gov.hhs.fha.nhinc.event.persistence.HibernateUtil.getSessionFactory();
```
vdmehta06, line 56: Same here - as below why we have full path instead why don't we have import statement?

christophermay07, line 56: Same reason as above.

vdmehta06:
Endorsing to merge without any further comments

PR #1243, opened 2016-02-02
Fix for java.util.ConcurrentModificationException

TabassumJafri:
1. ConnectionManagerCacheHelper: Modified mergeBusinessEntityServices() method to create and return a new BusinessEntity that holds merged enteries of both internatConnectionInfo and UddiConnectionInfo.
2. ConnectionManagerCache: Modified getAllBusinessEntities(), getBusinessEntity(), getBusinessEntityByHCID and getBusinessEntityByServiceName() to use the return value from ConnectionManagerCacheHelper's mergeBusinessEntityServices() method.

@christophermay07

### DIFF
--- a/Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/connectmgr/ConnectionManagerCache.java
+++ b/Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/connectmgr/ConnectionManagerCache.java
@@ -297,8 +297,8 @@ public class ConnectionManagerCache implements ConnectionManager {
             if (NullChecker.isNotNullish(homeCommunityId)) {
                 BusinessEntity oExistingEntity = helper.extractBusinessEntity(allEntities, homeCommunityId);
                 if (oExistingEntity != null) {
-                    helper.mergeBusinessEntityServices(oExistingEntity, uddiEntity);
-                    helper.replaceBusinessEntity(allEntities, oExistingEntity);
+                    helper.replaceBusinessEntity(allEntities, helper.mergeBusinessEntityServices(
+                        oExistingEntity, uddiEntity));
                 } else {
                     allEntities.add(uddiEntity);
                 }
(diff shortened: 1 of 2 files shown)


TabassumJafri:
Assigned for review: @danielrf123

danielrf123:
I have reviewed this code and am OK for a merge.

christophermay07:
I will continue looking into this, but wanted to ask if you've already looked into the following:

https://github.com/CONNECT-Solution/CONNECT/pull/1243/files#diff-6a69ac106d310e3e2980cbf5bdd6c893R973

After your change, the internalEntity object used in the location above may have fewer entries than before.  Can this cause updateInternalServiceUrl to return different data than before?  If so, is there any wider impact on CONNECT?

danielrf123:
I'm not seeing how the updates to mergeBusinessEntityServices() would affect what is done in updateInternalServiceUrl().  All places where mergeBusinessEntityServices() is used (getAllBusinessEntities(), getBusinessEntity(String), getBusinessEntityByHCID(String), getBusinessEntityByServiceName(String)), are not called in there.  Are they invoked earlier, and could that cause the data in updateInternalServiceUrl() to return different data?

Also, I haven't digged too deep into where the BusinessEnitity.businessService is used and how it works, but wouldn't the new mergedEntity field end up having the same number of entries (line 71)?

christophermay07:
@danielrf123 The original and the modified function will return the same value, but the original function also modified internalEntity.  If that internalEntity is a pointer to a value in the ConnectionManagerCache's "m_hInternalConnectInfo" map, then there can potentially be side effects anywhere we get that value.

christophermay07:
Re: my original comment -- @TabassumJafri and I discussed offline and arrived at the following conclusion: since we are no longer modifying the Cache's ICI map, functions that rely on those business service entries will branch differently than before -- but in the end, all functions will continue to behave the same due to fallback functionality to grab UDDI entries where they do not exist in ICI.

christophermay07:

PR #1248, opened 2016-02-09
FHAC-846 Fix DQ passthrough auditing to be consistent

vdmehta06:
For Query for Documents service, audit logging sequence is changed in order to fix the Time Out issue. While testing for QD 15-20 times, I observed that Time Out issue for X12 Real time service twice in 15-20 time cycle, but I was not able to re-create the same issue every time.

PR Review: @alameluchidambaram @mhpnguyen


mhpnguyen:
I've reviewed and endorse for merging without any comments

alameluchidambaram:

PR #1265, opened 2016-04-21
fix remaining nullpointer issue in sonarqube

mhpnguyen:
Fix remaining "Null pointers should not be dereferenced in sonarqube". 
@danielrf123 @vdmehta06 : can you please review this?


danielrf123:
I am ok with the fix and endorse for merge.

vdmehta06:
Endorsing to merge without any comments

PR #1269, opened 2016-04-27
Resolved the two new TRUE/FALSE condition blocker issues

danielrf123:
Resolved the two new TRUE/FALSE condition blocker issues
@mhpnguyen and @TabassumJafri, can you please review?


mhpnguyen:
Endorse for merging without comments

TabassumJafri:
Reviewed and Good with merge

PR #1270, opened 2016-05-09
Deactivate glassfish and generate wildfly EAR by default.

supriyaGit:
When Connect builds the binaries (EAR), the system generate glassfish ear by default. Since we don't official support glassfish, we want the system to generate wildfly (jboss) ear by default unless users explicitly define glassfish profile.
@mhpnguyen and @TabassumJafri can you please review?

### DIFF
--- a/Product/Production/Deploy/pom.xml
+++ b/Product/Production/Deploy/pom.xml
@@ -10,11 +10,10 @@
     <name>CONNECT Deployment</name>
     <packaging>pom</packaging>
     <modelVersion>4.0.0</modelVersion>
-
-    <modules>
-        <module>glassfish</module>
-    </modules>
-
+		<modules>
+             <module>jboss7</module>
+         </modules>
+    
     <reporting>
         <plugins>
             <plugin>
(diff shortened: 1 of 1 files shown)


mhpnguyen:
add to whitelist

Product/Production/Deploy/pom.xml
```diff
@@ -824,12 +823,15 @@
...
-            <modules>
-                <module>jboss7</module>
-            </modules>
-        </profile>
-    </profiles>
+        <!-- Glassfish -->
+     	<profile>
+            <id>glassfish</id>
+			<activation>
+                <activeByDefault>true</activeByDefault>
```
TabassumJafri, line 830: Why we need activeByDefault as true in glassfish profile?

mhpnguyen, line 830: Since we don't want to break CI job for this PR, we will leave it as default. The next PR will remove this.

mhpnguyen:
Endorse for merge without comment

TabassumJafri:
Endorsing for merge

PR #1286, opened 2016-05-19
FHAC-462 Create a profile for these Manually executed SoapUI Testcases

supriyaGit:
Test cases are manually executed and the poms are not updated while executing mvn updateversions for Release version changes during post Release. 

Created "manual-testing" profile for all Manual Test Cases that listed in FHAC-462 so that the Release versions can be updated without ignoring these child poms. 

@mhpnguyen and @TabassumJafri can you please review?


mhpnguyen:
Endorse for merging without comments

TabassumJafri:
Endorsing for merge without any comments.

PR #1294, opened 2016-06-01
FHAC 1031: Fix tear down script for any jmx projects

mhpnguyen:
- Apply logic in teardown script to determine whether the system should switch from standard to passthrough mode.
  @TabassumJafri (optional) @danielrf123 @sailajaa : Can you please review this?


sailajaa:
I reviewed PR and good with changes. It will be good if we can format tear down scripts, but leaving it for developer choice.

mhpnguyen:
I will address format for tear down script.

danielrf123:
I have reviewed and endorse for merge after the comment above has been addressed.

danielrf123:
Starting the merge.

PR #1317, opened 2016-07-12
FHAC-965: fix sonarqube issue

mhpnguyen:
fix Exception handlers should preserve the original exception in sonaqube.

@TabassumJafri, @danielrf123, @sailajaa (Optional): Can you please review this?

### DIFF
--- a/Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/DirectAdapter.java
+++ b/Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/DirectAdapter.java
@@ -109,12 +109,12 @@ public abstract class DirectAdapter {
             urlString = oProxyHelper
                     .getAdapterEndPointFromConnectionManager(DIRECT_CONFIG_SERVICE_NAME);
         } catch (ConnectionManagerException ex) {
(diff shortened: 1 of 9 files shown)


TabassumJafri:
It is better to use exception.getLocallizedMessage  instead of exception.getMessage because of internalization. I have seen both version getting used in this PR. Can we decide and stick on one?

Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/DirectAdapter.java
```diff
@@ -109,12 +109,12 @@ private URL getUrl() {
             urlString = oProxyHelper
                     .getAdapterEndPointFromConnectionManager(DIRECT_CONFIG_SERVICE_NAME);
         } catch (ConnectionManagerException ex) {
-            LOG.error("Error reading directConfig Url cannot be found: " + ex.getMessage());
+            LOG.error("Error reading directConfig Url cannot be found: {}", ex.getLocalizedMessage(), ex);
```
danielrf123, line 112: I am OK with just using ex.getErrorMessage() than getLocalizedMessage(), but I do agree that we should try to stick with one method in our code for loggers.

TabassumJafri:
Good with merge without any further comments.

PR #1318, opened 2016-07-13
add default constructor since spring bean can not instantiate without it

mhpnguyen:
Spring can not instantiate bean classs without default constructor.  Create default constructor without default.
@TabassumJafri , @danielrf123 : can you please review this?


TabassumJafri:
Good with merge without any comments.

danielrf123:
I endorse for merge.

PR #1319, opened 2016-07-13
FHAC-967: Sonarqube reports 2 Direct Critical issues: Instance methods should not write to static fields

danielrf123:
Removed the instanceof operator from line 154 to solve a major SonarQube finding.

@TabassumJafri @mhpnguyen please review.

### DIFF
--- a/Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/DirectAdapter.java
+++ b/Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/DirectAdapter.java
@@ -59,7 +59,7 @@ public abstract class DirectAdapter {
     static final String DIRECT_CONFIG_SERVICE_NAME = "directconfig";
 
     private final MailSender externalMailSender;
-    private static SmtpAgent smtpAgent = null;
+    private SmtpAgent smtpAgent = null;
     private final DirectEventLogger directEventLogger;
 
     /**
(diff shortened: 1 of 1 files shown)


Product/Production/Services/DirectCore/src/main/java/gov/hhs/fha/nhinc/direct/DirectAdapter.java
```diff
@@ -140,47 +133,42 @@ private MessageProcessResult processAsDirectMessage(MimeMessage mimeMessage) {
...
+     * certificates, anchors, domains, settings etc. The default cache refresh time is 5 minutes and can be overridden
+     * using the gateway property AgentSettingsCacheRefreshTime.
      *
      * @param url the directConfig Service url String converted into URL format.
      * @return SmtpAgent SmtpAgent created to process Direct Messages.
      */
     protected SmtpAgent getSmtpAgent(URL url) {
 
-        if ((smtpAgent != null) && smtpAgent instanceof SmtpAgent) {
+        if (smtpAgent != null) {
```
mhpnguyen, line 154: possible to remove null check and use instanceof: 
Please see the link for more information: http://stackoverflow.com/questions/2950319/is-null-check-needed-before-calling-instanceof

danielrf123, line 154: instanceOf is not needed because the smtpAgent member is always an instance of SmtpAgent.

TabassumJafri:
Good with merge without any comments.

mhpnguyen:
Endorse for merging without any comments.

PR #1322, opened 2016-07-28
FHAC-1098: Classloader changes for Wildfly/Jboss

danielrf123:
The deployment descriptor for the Jboss deployment needs to exclude the implicit dependency for org.jboss.logging to get passed the classNotFound exception during deployment.  With this fix, we will not need to include the jboss-logging-3.3.0.Final.jar in the wildfly server that is launched in ValidationSuite.

@mhpnguyen @TabassumJafri can you please review?


mhpnguyen:
endorse for merging without comments

TabassumJafri:
Endorsing for merge without any comments

PR #1332, opened 2016-08-04
FHAC-1080: Add messageID prefix to present message IDs in getAssertion()

danielrf123:
Document Submission Deferred Response does not have prefix included in the BEGIN_OUTBOUND_PROCESSING messageID.  This is happening because the assertion being read in BaseService.getAssertion() already includes the messageID, so no mesageID is "generated", which is where the formatting to include the prefix to the ID is done at.

@TabassumJafri @mhpnguyen please review.

### DIFF
--- a/Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/messaging/server/BaseService.java
+++ b/Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/messaging/server/BaseService.java
@@ -32,6 +32,7 @@ import gov.hhs.fha.nhinc.cxf.extraction.SAML2AssertionExtractor;
 import gov.hhs.fha.nhinc.nhinclib.NhincConstants;
 import gov.hhs.fha.nhinc.nhinclib.NullChecker;
 import gov.hhs.fha.nhinc.util.HomeCommunityMap;
+import gov.hhs.fha.nhinc.wsa.WSAHeaderHelper;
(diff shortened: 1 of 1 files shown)


TabassumJafri:
Endorsing for merge without any comments

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/messaging/server/BaseService.java
```diff
@@ -64,8 +66,13 @@ protected AssertionType getAssertion(WebServiceContext context, AssertionType as
         }
 
         // Extract the message id value from the WS-Addressing Header and place it in the Assertion Class
-        if (assertion != null && NullChecker.isNullish(assertion.getMessageId())) {
-            assertion.setMessageId(extractor.getOrCreateAsyncMessageId(context));
+        if (assertion != null) {
```
mhpnguyen, line 69: [need to confirm]no need to check null b/c the obj is not null when reaching that block.

danielrf123, line 69: Could be possible that SAML2AssertionExtractor.getInstance().extractSamlAssertion(context) returns null (line 63).  SonarQube should have complained if its not possible to be null in line 69.

mhpnguyen:
endorse for merging without comments

PR #1333, opened 2016-08-11
fix restore payload element in x12 after audit

mhpnguyen:
Since we don't audit payload elememnt, we set payload element to null which causes the original value to change.  Therefore, we need to assign to a tempoary variable to hold  references and reassign after marshing.

@TabassumJafri, @danielrf123, @sailajaa(Optional); Can you please review this?


sailajaa:
I reviewed and good with merge without comments.

TabassumJafri:
Reviewed and good with merge without any comments.

danielrf123:
Reviewed and good with merge.

PR #1335, opened 2016-08-11
FHAC-1164: disable redundant groovy script

mhpnguyen:
Clean up soapUI project and disable redundant groovy steps which causes project to fail

@sailajaa @danielrf123 , @TabassumJafri (optional); can you please review this


sailajaa:
I think we do not need disabled "Copy uddiConnectionInfo" groovy script step . Can we remove that.

danielrf123:
Endorse for merge with no comments.

sailajaa:
I reviewed and good with PR merge.

PR #1339, opened 2016-09-29
Fix JDK 8 build

christophermay07:
JUnit and JMock updates to support building CONNECT on JDK 8.

### DIFF
--- a/Product/pom.xml
+++ b/Product/pom.xml
@@ -559,7 +559,7 @@
         <groupId>org.javassist</groupId>
         <artifactId>javassist</artifactId>
         <version>3.20.0-GA</version>
-        <scope>provided</scope>
+        <scope>runtime</scope>
       </dependency>
 
       <!-- Logging -->
--- a/Product/pom.xml
+++ b/Product/pom.xml
@@ -556,9 +556,9 @@
         <scope>provided</scope>
       </dependency>
       <dependency>
-        <groupId>javassist</groupId>
+        <groupId>org.javassist</groupId>
         <artifactId>javassist</artifactId>
-        <version>3.4.GA</version>
+        <version>3.20.0-GA</version>
         <scope>provided</scope>
       </dependency>
 
@@ -1087,25 +1087,25 @@
       <dependency>
         <groupId>org.jmock</groupId>
         <artifactId>jmock</artifactId>
-        <version>2.5.1</version>
+        <version>2.8.2</version>
         <scope>test</scope>
       </dependency>
       <dependency>
         <groupId>org.jmock</groupId>
         <artifactId>jmock-legacy</artifactId>
-        <version>2.5.1</version>
+        <version>2.8.2</version>
         <scope>test</scope>
       </dependency>
       <dependency>
         <groupId>org.jmock</groupId>
         <artifactId>jmock-junit4</artifactId>
-        <version>2.5.1</version>
+        <version>2.8.2</version>
         <scope>test</scope>
       </dependency>
       <dependency>
         <groupId>org.mockito</groupId>
         <artifactId>mockito-all</artifactId>
-        <version>1.9.0</version>
+        <version>1.10.19</version>
         <scope>test</scope>
       </dependency>
       <dependency>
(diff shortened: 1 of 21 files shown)


mhpnguyen:
We have reviewed the PR and tested successful but will merge later after CXF upgrade.

christophermay07:
@mhpnguyen @jsmith2 PR has been rebased and built successfully on JDK8.  Build not tested on JDK7, and ValidationSuite has not been run on either JDK.

mhpnguyen:
@christophermay07 : The server gets the following error "[ERROR] Caused by: java.lang.NoClassDefFoundError: javassist/util/proxy/MethodFilter"  which causes ValidationSuite fails on Github Pull Request Validator job.  We will take a look at further.

christophermay07:
@mhpnguyen Looks like in the CXF 3 pull request, javassist (along with many other jars) were changed to "provided" scope.  Many of these jars are not necessarily provided by all supported application servers, in which case they should probably either be changed to runtime (if needed to compile/test) or removed (if not needed at all).  Some of the following may be correct as "provided", but here is list of dependencies that I strongly suspect are incorrectly scoped:

asm, asm-attrs, cglib-nodep, commons-dbcp, commons-net, commons-pool, cxf-rt-databinding-aegis, cxf-tools-common, cxf-tools-validator, cxf-tools-wsdlto-core, cxf-tools-wsdlto-frontend-jaxws, enunciate-core-annotations, esapi, FastInfoset, jackson-core-asl, jackson-jaxrs, jackson-mapper-asl, javassist, jcip-annotations, jcl-over-slf4j, json, jul-to-slf4j, myfaces-bundle, not-yet-commons-ssl, spring-jms, spring-messaging, spring-webmvc, xml-apis, xmlbeans, xmlpull, xpp3_min

Additionally, the scope of these dependencies should be double-checked at some point:

geronimo-javamail_1.4_spec, geronimo-jta_1.1_spec, jsr311-api, jandex, jsr173_api, stax-api, activation, xstream

Product/pom.xml
```diff
@@ -556,9 +556,9 @@
         <scope>provided</scope>
       </dependency>
       <dependency>
-        <groupId>javassist</groupId>
+        <groupId>org.javassist</groupId>
         <artifactId>javassist</artifactId>
-        <version>3.4.GA</version>
+        <version>3.20.0-GA</version>
```
mhpnguyen: Since javassist(3.20.0-GA) is loaded by other dependency, we should not exclude this jar as CONNECT requires.  Revert this javassist to 3.4.GA will make ValidationSuite pass in both jdk7 and jdk8 (1.8.0_66).

mhpnguyen (changes requested):
please see in-line comments

christophermay07:
@mhpnguyen I didn't see your comment until now, but I was able to run Validation Suite by changing the scope, not the version, of javassist.  Please let me know how you'd like me to proceed.

mhpnguyen:
@christophermay07 You are right.  CONNECT is only required to have javassit 3.20.0-GA.jar.  In 4.7, we saw both javaassit 3.4.GA and 3.20.0-GA jar in WAS.  You can either provide runtime scope which it will package inside CONNECT ear or let other dependency load it.  Some of the jars that you mentions are not required in CONNECT ear(which is why we use provided scope-kind of hack to exclude global).  We use dependecy-tree to check which jar are packed when building the ear.  We can remove them by using exclude or delete but some other dependencies load them again(it takes too much time to using exclude ).    We tested successful non-direct connect and direct-connect in all app servers(wildfly/eap7/weblogic/was) with dramatic reduce number of jars.

I tested your new change and it works for both jdk7/jdk8.  We will definitely look into those jars again if we have time.

mhpnguyen (approved):

christophermay07:
@mhpnguyen Thank you for the explanation -- if you're okay with the single scope-change for javassist in this PR, I think any further scope-change discussions belong in a different ticket/PR.

I don't consider this a critical issue, but definitely keep it on the radar so it doesn't get kicked too far down the road.

PR #1342, opened 2016-11-02
Fixed Large Payload file lock

TabassumJafri:
Created a Custom DataSource class that's is extended from FileDataSource. The main purpose of this class is to return the same instance of InputStream and OutputStream for a given object. User can call inputStream.close to close the underlying IinputStream. The DataHandler class is then created by providing the Custom  DataSource class and user is now able to release the file lock by calling close on the input stream. 

@mhpnguyen


mhpnguyen (approved):
No comment for this PR.  Approve for merging

PR #1354, opened 2017-01-25
FHAC-1331: fix jcl classloader issue in websphere for direct

mhpnguyen:
Fix log issue due to multiple jar presents in websphere classpath. 
@jsmith2 @mikhelamdex  @sjarral112: can you please review?

### DIFF
--- a/Product/Production/Adapters/General/CONNECTAdminGUI/pom.xml
+++ b/Product/Production/Adapters/General/CONNECTAdminGUI/pom.xml
@@ -11,10 +11,6 @@
     <packaging>war</packaging>
     <modelVersion>4.0.0</modelVersion>
 
-    <properties>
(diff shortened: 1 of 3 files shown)


christophermay07:
With the exclusion of the JCL to SLF4J bridge as a transitive dependency, do you know if this has any side effect on what goes to the logs? (e.g., will third party libraries that use JCL statements still output to the CONNECT logs?)

Also, how does this affect the effective POM for non-WebSphere builds?

mhpnguyen:
We test multiple servers (eap 7, WAS, weblogic,widfly) and log is fine.  We will test further more.

jsmith2:
I approve this PR with no suggested changes.

jsmith2 (approved):

PR #1366, opened 2017-02-14
Update dependency libraries to exclude unwanted jar

mhpnguyen:
Fix to exclude unwanted jars. Reduce 173 jars(4.7 release) to 155 for all services. 
@jsmith2(optional), @sjarral112 , @mikhelamdex : Can you please review?


sjarral112 (commented):
Approval for endorsed to merge

PR #1376, opened 2017-03-07
FHAC-1363: Fix ValidationSuite Tests fail when offline

mhpnguyen:
When internet cuts off,  CONNECt is unable to download ws-addr.xsd from public which cause ValidationSuite Tests fail.  The jax-ws-catalog.xml will allow CXF to reroute the public url to local schema when offline.  This also requires xmlresolver.jar to parse jax-ws-catalog.xml.
@TabassumJafri , @sjarral112: Can you please review?


sjarral112 (approved):
Approved to be merged !

TabassumJafri (approved):
good with merge

PR #1402, opened 2017-05-10
Remove spring-message version range

mhpnguyen:
Remove spring-message version range since it causes an issue to retrieve "io.projectreactor:reactor-bom:pom:Bismuth-M1" artifact.


TabassumJafri (approved):

PR #1431, opened 2017-07-18
FHAC-1525b Added quick setup files for WildFly and JBoss

sovanncgi:
Added wildfly_quick_setup.zip and jboss_quick_setup.zip


TabassumJafri (approved):

trantang (approved):
reviewed

PR #1442, opened 2017-08-21
CONN-1751 conditions optional

trantang:
Update authz decision statement conditions to be optional and do not need processing

### DIFF
--- a/Product/Production/Common/CONNECTCoreLib/src/test/java/gov/hhs/fha/nhinc/callback/opensaml/HOKSAMLAssertionBuilderTest.java
+++ b/Product/Production/Common/CONNECTCoreLib/src/test/java/gov/hhs/fha/nhinc/callback/opensaml/HOKSAMLAssertionBuilderTest.java
@@ -29,7 +29,6 @@ package gov.hhs.fha.nhinc.callback.opensaml;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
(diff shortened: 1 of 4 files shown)


Product/Production/Common/CONNECTCoreLib/src/test/java/gov/hhs/fha/nhinc/callback/opensaml/HOKSAMLAssertionBuilderTest.java
```diff
@@ -369,10 +370,7 @@ public void testCreateAuthenticationDecisionStatements() throws PropertyAccessEx
         assertEquals(issuer.getFormat(), SAMLAssertionBuilder.X509_NAME_ID);
 
         final Conditions conditions = assertion.getConditions();
-        assertTrue(
-            beforeCreation.isBefore(conditions.getNotBefore()) || beforeCreation.isEqual(conditions.getNotBefore()));
-        assertTrue(beforeCreation.isBefore(conditions.getNotOnOrAfter())
-            || beforeCreation.isEqual(conditions.getNotOnOrAfter()));
+        assertNull(conditions);
```
TabassumJafri: We are just asserting on null here. Could we have some more meaningful values to test for?

TabassumJafri:
Not specific to this PR however I see there are some constants present in both SAMLConstants and NhincConstants. Can we port all saml related constants in SAMLConstants? For instance, we have these 2 values defined in both classes:
   String EVIDENCE_CONDITION_NOT_BEFORE_PROP = "evidenceConditionNotBefore";
   String EVIDENCE_CONDITION_NOT_AFTER_PROP = "evidenceConditionNotAfter";

TabassumJafri (changes requested):

Product/Production/Common/CONNECTCoreLib/src/test/java/gov/hhs/fha/nhinc/callback/opensaml/HOKSAMLAssertionBuilderTest.java
trantang: based on the given the values, the conditions is null

TabassumJafri (approved):

ptambellini01 (approved):

PR #1447, opened 2017-08-24
Database less CONNECT in passthru mode

TabassumJafri:
* Set the hibernate beans to lazy initialize
* Removed resource-refs from web.xml, jboss-web.xml and ibm-web-bnd.xml
* Removed unused jboss-web.xml and ibm-web-bnd.xml
* Removed AuditEJB initialization from the InitServlet class


Product/Production/Common/CONNECTCommonWeb/src/main/webapp/WEB-INF/web.xml
```diff
@@ -30,10 +30,10 @@
         <welcome-file>index.jsp</welcome-file>
     </welcome-file-list>
 
-    <resource-ref>
+    <!--resource-ref>
```
mhpnguyen: (Optional): Remove instead of comment out

jsmith2 (approved):

mhpnguyen (approved):

PR #1457, opened 2017-09-08
CONN-17771 Correct NHINSAMLAttributeTest module

mhpnguyen:
Correct misspelling NHINSAMLAttributeTest module in Regression Test


sovanncgi (approved):
Confirmed module name changed correctly

PR #1463, opened 2017-09-13
CONN-1768: Add NHINSAMLSubjectConfirmationTest into RegressionSuite

mhpnguyen:
- Add NHINSAMLSubjectConfirmationTest to cover multiple assertion subject confirmation methods feature
@pvenkatakrishnan1 , @trantang (Optional):


sovanncgi:
Two things:

- Move the test suite to New Features and be sure to update pom.xml as needed (and everywhere it's needed)
- Test needs to run in both standard and passthru mode (will move test into bimodal folder after 5.1 release)

jsmith2:
@sovanncgi  Is there a reason you want it run in both modes.  The same logic should be hit regardless.  Also, I don't think we should use the new feature folder anymore.  we've still got the keepaliveheader test in there from last release.  Just creates one more step to move later.  Better to inventory current release tests on our release testing page and make sure they are tested again later (especially if they are manual).

jsmith2 (approved):

sovanncgi:
I'm okay with only running in passthrough and keeping it in the passthrough folder

PR #1476, opened 2017-09-26
CONN-1876b add database-fields to patient-inputs

trantang:
QA-requested that database fields be added to the user-inputs

Minh is optional for review


ptambellini01 (approved):

TabassumJafri (approved):

PR #1483, opened 2017-10-05
CONN-1739 added profile to execute sql scripts to maven build

ptambellini01:


Product/Production/Deploy/pom.xml
```diff
@@ -11,6 +11,14 @@
   <packaging>pom</packaging>
   <modelVersion>4.0.0</modelVersion>
   
+    <properties>
+        <db.host>localhost</db.host>
+        <db.password>nhincpass</db.password>
```
pvenkatakrishnan1: Is it better to keep the username password in different settings file rather than in the pom?

ptambellini01: This is already defined in the SoapUI_Test pom. So to be consistent I had to place it here.

Product/Production/Deploy/pom.xml
```diff
@@ -479,7 +487,50 @@
         </plugins>
       </build>
     </profile>
-
+    <profile>
+	     <id>SBScripts</id>
```
mhpnguyen: maybe use different name for relevant meaning

Product/Production/Deploy/pom.xml
```diff
@@ -11,6 +11,14 @@
   <packaging>pom</packaging>
   <modelVersion>4.0.0</modelVersion>
   
+    <properties>
+        <db.host>localhost</db.host>
```
TabassumJafri: Define db.host and other db config params in pom at CONNECT folder. Remove it from both SoapUI_Test and Deploy.

Product/Production/Deploy/pom.xml
```diff
@@ -479,7 +487,50 @@
...
+          <plugins>
+            <plugin>
+              <groupId>org.codehaus.mojo</groupId>
+              <artifactId>sql-maven-plugin</artifactId>
+              <version>1.5</version>
+            <dependencies>
+              <dependency>
+                <groupId>mysql</groupId>
+                <artifactId>mysql-connector-java</artifactId>
+                <version>5.1.10</version>
```
mhpnguyen: Define version in connect pom and use place holder to display similar like this<version>${mysqldriver.version}</version>

Product/Production/Deploy/pom.xml
```diff
@@ -479,7 +487,50 @@
...
+            <executions>
+              <execution>
+                <phase>process-test-resources</phase>
+                <goals>
+                  <goal>execute</goal>
+                </goals>
+                <configuration>
+                  <autocommit>true</autocommit>
+                  <srcFiles>
+                    <srcFile>../../../DBScripts/nhincdb/dropall.sql</srcFile>
```
TabassumJafri, line 515: Also define path to dbscripts in CONNECT pom.xml

TabassumJafri (changes requested):

Product/Production/Deploy/pom.xml
ptambellini01: Cannot establish an absolute path. Will go ahead with the present path assignment

TabassumJafri (approved):

mhpnguyen (approved):

PR #1490, opened 2017-10-19
fix dead link and non-function x buttons

mhpnguyen:
Remove or close primeface dialog modal when exception happens.  This will allow users to recover to login again instead of block user out.


TabassumJafri (commented):
Reviewed .. will approve once the PR Validator is run

TabassumJafri (approved):

pvenkatakrishnan1 (commented):
Looks good

PR #1503, opened 2017-10-27
QA Fix

TabassumJafri:


Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/exchange/transform/uddi/UDDITransform.java
```diff
@@ -215,50 +215,50 @@ private static void buildTargetRegions(BusinessEntity entity, OrganizationType o
         }
     }
 
-    private static List<String> buildEmails(Contact contact) {
-        List<String> emails = null;
+    private static void buildEmails(Contact contact, ContactType cType) {
+        List<String> emails;
         if (contact != null && CollectionUtils.isNotEmpty(contact.getEmail())) {
```
mhpnguyen, line 220: Optional:  contact will never be null in this case since for loop above guarantee

trantang (approved):

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/exchange/transform/uddi/UDDITransform.java
```diff
@@ -215,50 +215,50 @@ private static void buildTargetRegions(BusinessEntity entity, OrganizationType o
...
             emails = new ArrayList<>();
             for (Email obj : contact.getEmail()) {
                 emails.add(obj.getValue());
             }
+            cType.getEmail().addAll(emails);
```
mhpnguyen, line 225: Question: wonder what happen at runtime if the list is empty

TabassumJafri: you will have an empty tag <email/>

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/exchange/transform/uddi/UDDITransform.java
TabassumJafri: contact is an element in the list. It could be null and that's why a null check.

mhpnguyen (approved):

PR #1504, opened 2017-11-01
CONN-1924 Updating ACP attributes to be the correct type

jsmith2:


Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/callback/opensaml/HOKSAMLAssertionBuilder.java
```diff
@@ -779,19 +786,31 @@ private static DateTime setEndValidTime(DateTime endValidTimeArg, DateTime issue
...
-                            SamlConstants.URI_NAME_FORMAT, Arrays.asList(iacp));
+                            SamlConstants.URI_NAME_FORMAT);
+            iacpAttribute.getAttributeValues().add(createURI(iacp));
             statements
                     .addAll(OpenSAML2ComponentBuilder.getInstance().createAttributeStatement(Arrays.asList(iacpAttribute)));
         }
         return statements;
     }
+    
+    private XMLObject createURI(String value) {
```
mhpnguyen: Why do we create the method in this class instead of OpenSAML2ComponentBuilder?  OpenSAML2ComponentBuilder is probably the right class to handle all creation saml attribute statement and value

jsmith2: Updating.  Will submit another commit to PR.

pvenkatakrishnan1 (approved):
Still learning the code base

mhpnguyen (approved):

PR #1515, opened 2017-11-14
Transformed uddi and ici file into new exchange schema for CONN-1923

TabassumJafri:


trantang (approved):
dont see anything wrong

sovanncgi (approved):

PR #1519, opened 2017-11-21
Fixed regressionSuite. These testcase reuired soapu ui fixes only

TabassumJafri:


mhpnguyen (approved):

trantang (approved):
i don't see anything wrong

PR #1549, opened 2017-12-12
CONN-1955 --> Fields in a "Serializable" class should either be transient or serializable

TabassumJafri:


mhpnguyen (approved):

trantang (approved):
did see anything wrong

PR #1584, opened 2018-01-24
Added final block to release DAO lock and the subsequenmt testcases

TabassumJafri:
Added final block to release DAO lock and the subsequenmt testcases


trantang (changes requested):
Request change for Test-methods

trantang (approved):

pvenkatakrishnan1 (approved):

PR #1629, opened 2018-04-04
CONN-2139 Test Data Document Correction

ptambellini01:


Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/services/LoadTestDataDBServiceImpl.java
```diff
@@ -73,7 +73,15 @@
 
     @Override
     public boolean deletePatient(Patient patient) {
-        return patientDAO.deleteTransaction(patient);
+        boolean actionResult;
+        List<Document> delDocument = documentDAO.findAllBy(patient.getPatientId());
+        actionResult = patientDAO.deleteTransaction(patient);
+        if (actionResult) {
+            for (int i = 0; i < delDocument.size(); i++) {
```
TabassumJafri: What happens if a patient is deleted however you get an exception while deleting a document? I think this whole process, that is deleting a patient and associated documents, should be one transaction.

TabassumJafri (changes requested):

plobre (approved):

TabassumJafri (approved):

PR #1630, opened 2018-04-06
fix PD secure and unsecure to use correct CONNECT Client factory

mhpnguyen:
Fix PD adapter to use correct service regarding secure and unsecure.

### DIFF
--- a/Product/Production/Services/PatientDiscoveryCore/src/main/java/gov/hhs/fha/nhinc/patientdiscovery/adapter/proxy/AdapterPatientDiscoveryProxyWebServiceHelper.java
+++ b/Product/Production/Services/PatientDiscoveryCore/src/main/java/gov/hhs/fha/nhinc/patientdiscovery/adapter/proxy/AdapterPatientDiscoveryProxyWebServiceHelper.java
@@ -1,7 +1,7 @@
 /*
  * Copyright (c) 2009-2018, United States Government, as represented by the Secretary of Health and Human Services.
  * All rights reserved.
- * 
+ *
  * Redistribution and use in source and binary forms, with or without
  * modification, are permitted provided that the following conditions are met:
  *     * Redistributions of source code must retain the above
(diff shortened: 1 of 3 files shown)


Product/Production/Services/PatientDiscoveryCore/src/main/java/gov/hhs/fha/nhinc/patientdiscovery/adapter/proxy/AdapterPatientDiscoveryProxyWebServiceHelper.java
```diff
@@ -73,12 +74,20 @@ public PRPAIN201306UV02 formatGatewayResponsePRPAIN201305UV02(PRPAIN201305UV02 b
...
                     request.setNhinTargetCommunities(null);
 
-                    ServicePortDescriptor<AdapterPatientDiscoverySecuredPortType> portDescriptor = new AdapterPatientDiscoverySecuredServicePortDescriptor();
-                    CONNECTClient<AdapterPatientDiscoverySecuredPortType> client = CONNECTClientFactory.getInstance()
-                        .getCONNECTClientSecured(portDescriptor, url, assertion);
-
-                    response = (PRPAIN201306UV02) client.invokePort(AdapterPatientDiscoverySecuredPortType.class,
+                    ServicePortDescriptor portDescriptor;
+                    CONNECTClient client;
+                    if (NhincConstants.ADAPTER_PATIENT_DISCOVERY_SECURED_SERVICE_NAME.equals(sServiceName)) {
```
jsmith2: Suggestion, pull out the added logic into a new method:  buildPortDescriptor(String sServiceName), that will pull out the added if/else statement.  Also rename this helper method.  Should be something like sendToPRPAIN201306UV02Adapter(...).  The current name "formatGatewayResponsePRPAIN201305UV02" isn't really what it does.  There is no formatting.

jsmith2 (changes requested):

TabassumJafri (commented):
Reviewed and will approve once the review comments are addressed

TabassumJafri (approved):

jsmith2 (approved):

PR #1636, opened 2018-04-19
Added adminws profile to build ConfigAdmin module

TabassumJafri:


trantang (approved):

Product/Production/Gateway/ConfigAdmin/src/main/webapp/WEB-INF/glassfish-web.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mhpnguyen, line 1: Optional: Since we don't support glassfish, should we remove it?

mhpnguyen (approved):

PR #1651, opened 2018-05-15
Merge CONN_2161 solution from Integration into 5.1.2

mhpnguyen:


Product/Production/Common/CONNECTCoreLib/src/test/java/gov/hhs/fha/nhinc/docrepository/adapter/dao/DocumentDaoIntTest.java
```diff
@@ -0,0 +1,181 @@
+/*
+ * Copyright (c) 2009-2018, United States Government, as represented by the Secretary of Health and Human Services.
+ * All rights reserved.
```
TabassumJafri, line 3: this class needs to be removed. This is new functionality for DS adapter

TabassumJafri (changes requested):

PR #1662, opened 2018-05-23
added overrides exchange type to exchange manager

TabassumJafri:
Also updated ExchangeScheduler to use Long object instead of primitive

### DIFF
--- a/Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/exchangemgr/ExchangeManager.java
+++ b/Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/exchangemgr/ExchangeManager.java
@@ -522,7 +522,7 @@ public class ExchangeManager extends AbstractExchangeManager<UDDI_SPEC_VERSION> 
             hcidWithPrefix, overrideHcid);
     }
 
-    private boolean isValidExchangeType(String type) {
+    private static boolean isValidExchangeType(String type) {
         for (EXCHANGE_TYPE exType : EXCHANGE_TYPE.values()) {
             if (StringUtils.equalsIgnoreCase(exType.getValue(), type)) {
                 return true;
--- a/Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/exchangemgr/ExchangeManager.java
+++ b/Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/exchangemgr/ExchangeManager.java
@@ -29,10 +29,12 @@ package gov.hhs.fha.nhinc.exchangemgr;
 import gov.hhs.fha.nhinc.connectmgr.persistance.dao.ExchangeInfoDAOFileImpl;
 import gov.hhs.fha.nhinc.exchange.ExchangeInfoType;
 import gov.hhs.fha.nhinc.exchange.ExchangeType;
+import gov.hhs.fha.nhinc.exchange.directory.EndpointConfigurationListType;
 import gov.hhs.fha.nhinc.exchange.directory.EndpointConfigurationType;
 import gov.hhs.fha.nhinc.exchange.directory.EndpointType;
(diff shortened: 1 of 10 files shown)


Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/exchangemgr/ExchangeManager.java
```diff
@@ -84,14 +87,10 @@ private void loadExchangeInfo() throws ExchangeManagerException {
...
-                        if (null != ex.getOrganizationList()
-                            && CollectionUtils.isNotEmpty(ex.getOrganizationList().getOrganization())
-                            && StringUtils.isNotEmpty(ex.getType()) && StringUtils.isNotEmpty(ex.getName())) {
-                            Map<String, OrganizationType> innerMap = new HashMap<>();
-                            for (OrganizationType org : ex.getOrganizationList().getOrganization()) {
-                                innerMap.put(org.getHcid(), org);
-                            }
-                            exCache.put(ex.getName(), innerMap);
+                        if (isOverrideExchange(ex)) {
+                            overrideExchange = ex;
```
plobre: Do we need to break here? What is the requirement / possibility of having multiple overrides defined? as it stands if we have more than one, the latest will stick and the previous ones will overwritten.

TabassumJafri: We discussed to have only one Override exchange type in exchangeInfo.xml.  ExchangeName is not important for an override exchange type  so  we cannot distinguish even if the user provides more than one override exchange. 
As discussed offline, i will add a warning when there are more than one override exchange type and the previous one is about to overwritten.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/exchangemgr/ExchangeManager.java
```diff
@@ -350,4 +330,169 @@ public boolean toggleIsDisabledFor(String forExchangeName) throws ExchangeManage
         }
         return bSave;
     }
+
+    private static boolean isOverrideExchange(ExchangeType exchange) {
+        return null != exchange.getOrganizationList()
+            && CollectionUtils.isNotEmpty(exchange.getOrganizationList().getOrganization())
+            && StringUtils.isNotEmpty(exchange.getType()) && StringUtils.equalsIgnoreCase(
```
mhpnguyen: Optional: No need to do StringUtils.isNotEmpty(exchange.getType()) since StringUtils.equalsIgnoreCase can handle null if exchange.getType()=null

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/exchangemgr/ExchangeManager.java
```diff
@@ -84,14 +87,10 @@ private void loadExchangeInfo() throws ExchangeManagerException {
...
                     for (ExchangeType ex : exInfo.getExchanges().getExchange()) {
-                        if (null != ex.getOrganizationList()
-                            && CollectionUtils.isNotEmpty(ex.getOrganizationList().getOrganization())
-                            && StringUtils.isNotEmpty(ex.getType()) && StringUtils.isNotEmpty(ex.getName())) {
-                            Map<String, OrganizationType> innerMap = new HashMap<>();
-                            for (OrganizationType org : ex.getOrganizationList().getOrganization()) {
-                                innerMap.put(org.getHcid(), org);
-                            }
-                            exCache.put(ex.getName(), innerMap);
+                        if (isOverrideExchange(ex)) {
```
mhpnguyen: Optional(Improvement): we should have logic to check if exchange name match to our pre-define list.  If doesn't match, we should log and ignore.

TabassumJafri:
retest this please

jsmith2 (approved):

mhpnguyen (approved):

PR #1667, opened 2018-05-31
Added fix for multiple service names

TabassumJafri:
Fix:
ExchangeManagerHelper -- modified the if check in getNhinServiceName method
NhincConstant - removed the IllegalArgumentException
PurposeOfUse -- removed the try catch block for IllegalArgumentException and added  if service is null, log and return

### DIFF
--- a/Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/callback/PurposeOfForDecider.java
+++ b/Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/callback/PurposeOfForDecider.java
@@ -62,7 +62,7 @@ public class PurposeOfForDecider {
 
         serviceName = NHIN_SERVICE_NAMES.fromValueString(action);
         if (null == serviceName) {
-            LOG.warn("Could not read purpose of / for action: service name is null");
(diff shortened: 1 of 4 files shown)


jsmith2 (approved):

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/callback/PurposeOfForDecider.java
```diff
@@ -60,11 +59,10 @@ public boolean isPurposeFor(CallbackProperties properties) {
...
         String action = properties.getAction();
-        try {
-            serviceName = NHIN_SERVICE_NAMES.fromValueString(action);
-        } catch (IllegalArgumentException ex) {
-            LOG.warn("Could not read purpose of / for action: {}", ex.getLocalizedMessage());
-            LOG.trace("Could not read purpose of / for action: {}", ex.getLocalizedMessage(), ex);
+
+        serviceName = NHIN_SERVICE_NAMES.fromValueString(action);
+        if (null == serviceName) {
+            LOG.warn("Could not read purpose of / for action: service name is null");
```
plobre: Can we include the action name that was not found for troubleshooting later on?

mhpnguyen (approved):

plobre (approved):

PR #1668, opened 2018-06-04
Timeout fixes

kginstructor:
This fixes the issue where timeouts aren't honored in certain circumstances, and forces the timeouts to be consistent across a much wider array of circumstances.


mhpnguyen:
ok to test

PR #1669, opened 2018-06-04
CONN-2199 fix AdminGUI Insert Load Test Data Document

ptambellini01:


trantang (approved):

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/docrepository/adapter/model/DocumentMetadata.java
```diff
@@ -538,6 +541,9 @@ public String getNewRepositoryUniqueId() {
     }
 
     public void setNewRepositoryUniqueId(String newRepositoryUniqueId) {
+        if (null != document && null == document.getRepositoryUniqueId()) {
```
TabassumJafri, line 544: Do we need to check if newRepositoryUniqueId is not null here?

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/docrepository/adapter/model/DocumentMetadata.java
```diff
@@ -218,6 +218,9 @@ public String getDocumentUniqueId() {
     }
 
     public void setDocumentUniqueId(String documentUniqueId) {
+        if (null != document) {
+            document.setDocumentUniqueId(documentUniqueId);
```
TabassumJafri, line 222: Do we need a not null check for documentUniqueId

ptambellini01: In this instance for the test data loader it's a required field in the admin gui. So it can't be null.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/docrepository/adapter/model/DocumentMetadata.java
trantang: the Repository-Unique-id is checked in the business-layer on the AdminGUI

trantang: currently the NewRepositoryUniqueId and RepositoryUniqueId are different when we first-initialize the Repository-Unique-id they are the same; newRepositoryUnuiqueId is part of the MetaData, and RepositoryUniqueId is part of document

mhpnguyen (approved):

TabassumJafri (approved):

PR #1673, opened 2018-06-11
Added org type, description and updated active to boolean value

TabassumJafri:


Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/exchange/transform/uddi/UDDITransform.java
```diff
@@ -298,4 +299,14 @@ private static String buildRole(Contact contact) {
         }
         return null;
     }
+
+    private static List<String> buildDescritpion(BusinessEntity entity) {
+        List<String> descList = new ArrayList<>();
+        if (null != entity && CollectionUtils.isNotEmpty(entity.getDescription())) {
```
plobre: I feel the isNotEmpty() check is not needed here. The for loop will take care of this case.

plobre:
retest this please

trantang (approved):

plobre (approved):

PR #1676, opened 2018-06-14
Conn 2196: DocumentUniqueID and RepoID are now populated when saving a document

plobre:
Makes it so that the adapter will now save the document unique ID as a generated sequence starting with "CONNECT0" and increment for each new row, and sets the repository ID to 2.2 by default, but can be overridden via the adapter.properties file.


mhpnguyen:
retest this please

plobre:
Fast forward / rebase didnt seem to work correctly with last push. Cleaning up commit history. Will push in new branch.

PR #1682, opened 2018-06-21
Converted Standard DDS to Passthrough DDS

TabassumJafri:


trantang (approved):

plobre (approved):

Product/Production/Gateway/DocumentDataSubmission_10/pom.xml
```diff
@@ -45,6 +45,19 @@
...
              <version>${parent.version}</version>
         </dependency>
+        <!-- Testing -->
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-all</artifactId>
```
plobre: Verify we need this. I didnt see any use of Mockito in your context tests.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/aspect/BaseEventAdviceDelegate.java
```diff
@@ -243,24 +238,36 @@ public void createNewEvent() {
...
+        try {
+            if (obj != null) {
+                if (obj instanceof AssertionType) {
+                    return (AssertionType) obj;
+                } else {
+                    return getAssertionType(obj.getClass().getDeclaredMethods(), obj);
+                }
+            }
+        } catch (IllegalAccessException | InvocationTargetException ex) {
+            LOG.info("Unable to extract AssertionType from Object", ex);
```
plobre: Can we log the object as well? At the very least the object's class. I would also consider this to be an "Error" level, not Info.

Product/Production/Gateway/DocumentDataSubmission_10/src/main/resources/documentdatasubmission10.properties
```diff
@@ -1,2 +1,2 @@
-docdatasubmission.inboundDocDataSubmission=gov.hhs.fha.nhinc.docdatasubmission.inbound.StandardInboundDocDataSubmission
-docdatasubmission.outboundDocDataSubmission=gov.hhs.fha.nhinc.docdatasubmission.outbound.StandardOutboundDocDataSubmission
\ No newline at end of file
+docdatasubmission.inboundDocDataSubmission=gov.hhs.fha.nhinc.docdatasubmission.inbound.PassthroughInboundDocDataSubmission
+docdatasubmission.outboundDocDataSubmission=gov.hhs.fha.nhinc.docdatasubmission.outbound.PassthroughInboundDocDataSubmission
```
plobre, line 2: @TabassumJafri this slipped by me. Slack is down, so using this as an alternative. Outbound is mapping to Inbound here in the properties file.

PR #1703, opened 2018-07-16
Update WSAddressingTest to assert on soapAction and WSAddressingAction in ServicePortDescriptor class

TabassumJafri:


trantang (approved):

Product/Production/Common/CONNECTCoreLib/src/test/java/gov/hhs/fha/nhinc/util/ReflectionHelper.java
```diff
@@ -0,0 +1,131 @@
...
+            return null;
+        }
+        if (fromPackage == null) {
+            LOG.debug("Unknown package.");
+            return null;
+        }
+        final List<Class<?>> rVal = new ArrayList<>();
+        try {
+            final List<Class<?>> targets = getAllClassesFromPackage(fromPackage, filter);
+            if (CollectionUtils.isNotEmpty(targets)) {
```
plobre, line 61: Just being picky, but it looks like "targets" is never going to be null, so this check can be completely removed. The for-loop will No Op when it is empty.

plobre (approved):

plobre:
retest this please

PR #1722, opened 2018-08-01
Enabled PLQ test in ValidationSuite and AuditLogging

TabassumJafri:
Also fixed the wsdl lookup for EventLogging Passthru


sovanncgi (commented):
I don't see any changes in event logging passthrough test (I see disabled=true removed for the other two)

sovanncgi (approved):

TabassumJafri:
PLQ test case was already enabled in Event-Logging Passthrough. The PLQ wsdl was pointing towards incorrect path. Fixed PLQ wsdl path.

PR #1725, opened 2018-08-02
Updated DB scripts

TabassumJafri:


trantang (approved):

trantang:
local-Patient and correlated-patient(same as local-patient) should not be inserted into the patient-correlated-identier

PR #1737, opened 2018-08-20
Conn 2250b

plobre:
CONNECT / CONNECT_integration have diverged. Pulling CONNECT into CONNECT_Integration


TabassumJafri:
Not going to pull in all those test commits into CONNECT_integration. They should not have been part of CONNECT branch in the first place.

PR #1758, opened 2018-09-13
View alerts on dashboard

vdmehta06:
Added system alerts ( Errors and Warnings ) on Dashboard panel. You should be able to view top 5 expired/expiring soon alerts. To see all the alerts you need to click "View all Alerts".

### DIFF
--- a/Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/StatusBean.java
+++ b/Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/StatusBean.java
@@ -220,7 +220,7 @@ public class StatusBean {
             LOG.error("Unable to get certificate details {}", e.getLocalizedMessage(), e);
             HelperUtil.addMessageError(GET_CERT_STORE_MSG, "Unable to fetch certificate details");
         }
-        return null;
+        return new ArrayList<>();
     }
 
     public static void addCertificates(List<CertificateDTO> input, List<CertificateDTO> output) {
--- a/Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/StatusBean.java
+++ b/Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/StatusBean.java
@@ -39,6 +39,8 @@ import gov.hhs.fha.nhinc.callback.opensaml.CertificateDTO;
 import gov.hhs.fha.nhinc.callback.opensaml.CertificateManagerException;
 import gov.hhs.fha.nhinc.event.model.EventCount;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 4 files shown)


Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/StatusBean.java
```diff
@@ -52,132 +60,184 @@
...
+	private final PingService pingService = new PingServiceImpl();
+	private List<AvailableService> services = new ArrayList<>();
+	private HorizontalBarChartModel eventBarChart;
+	private PieChartModel eventPieChart;
+	private StatusSnapshot snapshot;
+	private final CertificateManagerService certificateService;
+
+	private static final String KEY_STORE_MSG = "keyStoreMsg";
+	private static final Logger LOG = LoggerFactory.getLogger(StatusBean.class);
+	public List<CertificateDTO> certDTOs = new ArrayList<CertificateDTO>();
```
connectscagitbot: MAJOR Make certDTOs a static final constant or non-public and provide accessors if needed. [rule](http://192.168.3.101:9000/coding_rules#rule_key=squid%3AClassVariableVisibilityCheck)

Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/StatusBean.java
```diff
@@ -52,132 +60,180 @@
...
+	private final PingService pingService = new PingServiceImpl();
+	private List<AvailableService> services = new ArrayList<>();
+	private HorizontalBarChartModel eventBarChart;
+	private PieChartModel eventPieChart;
+	private StatusSnapshot snapshot;
+	private final CertificateManagerService certificateService;
+
+	private static final String KEY_STORE_MSG = "keyStoreMsg";
+	private static final Logger LOG = LoggerFactory.getLogger(StatusBean.class);
+	public List<CertificateDTO> certDTOs = new ArrayList<>();
```
connectscagitbot: MAJOR Make certDTOs a static final constant or non-public and provide accessors if needed. [rule](http://192.168.3.101:9000/coding_rules#rule_key=squid%3AClassVariableVisibilityCheck)

Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/StatusBean.java
```diff
@@ -52,132 +60,180 @@
...
+
+	public StatusBean() {
+		certificateService = new CertificateManagerServiceImpl();
```
TabassumJafri: Optional: don't need a constructor here, you can use the @PostConstruct method to achieve the same result

Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/StatusBean.java
```diff
@@ -52,132 +60,180 @@
...
+		return series;
+	}
+
+	private void initPieChart() {
+		eventPieChart = new PieChartModel();
+		refreshPieChart();
+	}
+
+
+	public List<CertificateDTO> getCertificateInfo() {
```
plobre: This method looks pretty expensive. Lots of "addAll"s. I don't see the need as to why we need expiredCerts and warnningCerts (Misspelled too, by the way). Perhaps consider using an ordered list instead? I would imagine we want to see the certificates with the least amount of valid days left first on the UI anyway - so the order on the list would be the amount of remaining days.

Another potential improvement is simply to iterate over the result of the certificate service directly in a private method which would include lines 218 through 226 here. That would eliminate the need to have to add all of the results (which is already a list) to a bigger list just for the sake of iteration.

Last but certainly not least - certDTO is not cleared when this method is run! You will get duplicated certificates each time this method is run...

Product/Production/Adapters/General/CONNECTAdminGUI/src/main/webapp/status.xhtml
```diff
@@ -47,9 +47,27 @@
                 <p:tabView id="tabview" effect="fade" effectDuration="normal" dynamic="true" activeIndex="0" styleClass="tabs-override tabs-ovr-dkblue" >
                     <p:tab title="Dashboard" closable="false">
                         <div class="content-block">
+                          
+                         <p:panel rendered="#{not empty statusBean.certDTOs}"  id="alertsPanel" widgetVar="alertsPanel" style="width: 100%; margin: 0 auto;">
+                          <h2  class="subsection-header">System Alerts and Notifications </h2>
+                         <div class="table-responsive table-orgs">
+                            <p:dataTable id="certificateTable" rows="5" widgetVar="certificateTable" var="cert" value="#{statusBean.certDTOs}" scrollable="TRUE" scrollHeight="200"  styleClass="table table-striped table-status" lazy="true">
+                                <p:column>
+                                    <h:outputText rendered="#{cert.expiryColorCoding eq 'RED'}" value="Certificate with alias #{cert.alias} expired. Use Certificate Manager to Update. " style="color: red" class="text-danger" />
```
plobre: I don't agree we should be saying "Use the certificate manager to update" on each and every entry. Can we say it once somewhere if we have any certs about to expire, or already expired?

Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/StatusBean.java
plobre: My personal opinion, but I'd rather it be initialized on line 69 like the rest of the services here, and getCertificateInfo() called in the initServices() method.

Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/StatusBean.java
plobre: Optional: while we are at it, could we rename this to just "certificates"?

Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/StatusBean.java
```diff
@@ -52,132 +60,180 @@
...
+					dto.setExpiryColorCoding(COLOR_CODING_CSS.YELLOW.toString());
+					warnningCerts.add(dto);
+				}
+			}
+			certDTOs.addAll(expiredCerts);
+			certDTOs.addAll(warnningCerts);
+
+		} catch (CertificateManagerException e) {
+			LOG.error("Unable to get certificate details {}", e.getLocalizedMessage(), e);
+			HelperUtil.addMessageError(KEY_STORE_MSG, "Unable to fetch certificate details");
```
trantang: you can removed this line; there no-messages element to broadcast the error

Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/StatusBean.java
```diff
@@ -58,15 +68,28 @@
     private HorizontalBarChartModel eventBarChart;
     private PieChartModel eventPieChart;
     private StatusSnapshot snapshot;
+    private final CertificateManagerService certificateService = new CertificateManagerServiceImpl();;
```
connectscagitbot: MINOR Remove this empty statement. [rule](http://192.168.3.101:9000/coding_rules#rule_key=squid%3AEmptyStatementUsageCheck)

Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/StatusBean.java
```diff
@@ -58,15 +68,28 @@
...
+
+    private static final String KEY_STORE_MSG = "keyStoreMsg";
+    private static final Logger LOG = LoggerFactory.getLogger(StatusBean.class);
+    public List<CertificateDTO> certificateList = null;
```
connectscagitbot: MAJOR Make certificateList a static final constant or non-public and provide accessors if needed. [rule](http://192.168.3.101:9000/coding_rules#rule_key=squid%3AClassVariableVisibilityCheck)

TabassumJafri (approved):

Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/StatusBean.java
```diff
@@ -180,4 +203,37 @@ private void initPieChart() {
...
+    public List<CertificateDTO> getCertificateInfo() {
+        try {
+            certificateList = new ArrayList<>();
+            addCertificates(certificateService.fetchKeyStores(), certificateList);
+            addCertificates(certificateService.fetchTrustStores(), certificateList);
+
+            Collections.sort(certificateList, new Comparator<CertificateDTO>() {
+                @Override
+                public int compare(CertificateDTO o1, CertificateDTO o2) {
+                    return o2.getExpiresInDays() > o1.getExpiresInDays() ? -1
```
plobre: I see what you're doing here, but it might be simpler ( and *much* easier to read ) to subtract the number of days from one another. Since Comparator doesn't need to explicitly have -1, 0, or 1, we can subtract the days which will give us a positive or negative number, or 0 if they are equivalent. Thus all 3 cases are handled.

Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/StatusBean.java
```diff
@@ -180,4 +203,37 @@ private void initPieChart() {
...
+                            : o2.getExpiresInDays() == o1.getExpiresInDays() ? 0 : 1;
+                }
+            });
+        } catch (CertificateManagerException e) {
+            LOG.error("Unable to get certificate details {}", e.getLocalizedMessage(), e);
```
plobre: Should we return an empty list here if the method fails? As it stands, we could technically get a partially populated list (also unordered) if fetching the Keystore is successful, but the Trust store throws the exception.

Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/StatusBean.java
vdmehta06: why do you want to subtract and get the number of days since the operator( > or = ) can get the result?

plobre: Having a ? operator nested inside of another ? operator is difficult to read in my opinion, and it appears to be only for the sake of explicitly returning -1, 0, or 1, which isn't necessary

Subtracting the days remaining is much simpler to read and will satisfy the requirements.
For example, if one object evaluated to 15, and another evaluated to 10, we would have the following situations:
Less than: 10 - 15 = -5
Greater than: 15 - 10 = 5
If we had 2 objects with similar dates: 10-10 = 0


Since the comparator only uses the sign of the integer for sorting, we don't have to convert the 5 and -5 to 1 and -1 respectively, and all 3 cases are covered with respect to the required contract for comparators leaving us with a very nice and simple implementation:

`return o1.getExpiresInDays() - o2.getExpiresInDays()`

vdmehta06: ok sure, addressing this.

Product/Production/Adapters/General/CONNECTAdminGUI/src/main/java/gov/hhs/fha/nhinc/admingui/managed/StatusBean.java
vdmehta06: sure, addressing with next PR.

plobre (approved):

trantang (approved):

ptambellini01 (approved):

PR #1782, opened 2018-10-16
CONN-2301: Release resource in final block

mhpnguyen:
Release system resource in finally block to clear any resource even exceptions occur

### DIFF
--- a/Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/callback/opensaml/CertificateManagerImpl.java
+++ b/Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/callback/opensaml/CertificateManagerImpl.java
@@ -38,7 +38,6 @@ import java.io.ByteArrayOutputStream;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.IOException;
-import java.io.OutputStream;
 import java.security.KeyStore;
 import java.security.KeyStore.PrivateKeyEntry;
 import java.security.KeyStoreException;
@@ -54,7 +53,6 @@ import java.util.Map;
 import javax.activation.DataHandler;
 import org.apache.commons.io.IOUtils;
 import org.apache.commons.lang.StringUtils;
-import org.cryptacular.util.StreamUtil;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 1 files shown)


vdmehta06 (approved):

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/callback/opensaml/CertificateManagerImpl.java
```diff
@@ -219,9 +217,10 @@ private static FileOutputStream updateCertEntry(final String oldAlias, final Str
             tstore.setCertificateEntry(newAlias, certificate);
             tstore.store(os, passkey.toCharArray());
         } catch (final IOException | NoSuchAlgorithmException | CertificateException | KeyStoreException ex) {
-            IOUtils.closeQuietly(os);
             LOG.error("Unable to update the Certifiate: ", ex.getLocalizedMessage(), ex);
             throw new CertificateManagerException(ex.getMessage(), ex);
+        } finally {
+            IOUtils.closeQuietly(os);
```
plobre: Unsure if we can make this change. We are returning the actual stream in the method. Are we doing anything with the stream after this method returns, and if we are, would closing it affect the implementation?

mhpnguyen: nice catch.  this method uses by updateCertificate.  We will close stream in updateCertificate instead.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/callback/opensaml/CertificateManagerImpl.java
```diff
@@ -452,9 +451,14 @@ private void storeCert() throws IOException, KeyStoreException, CertificateExcep
             storeType = JKS_TYPE;
         }
         if (password != null && !PKCS11_TYPE.equalsIgnoreCase(storeType)) {
-            OutputStream os = new FileOutputStream(storeLoc);
-            trustStore.store(os, password.toCharArray());
-            StreamUtil.closeStream(os);
+            FileOutputStream os = null;
```
TabassumJafri: Optional: Can we use Try-with-resource here

TabassumJafri (approved):

plobre (approved):

PR #1831, opened 2018-12-14
Add Support Multiple SHA.

mhpnguyen:
Do not merge this PR

- Enable feature to support multiple SHA for backend:
-- Add configurableSHA entry in gateway.property as following: SHA1,SHA256,SHA512.  
- Enable to support multiple SHA for webservice client:
-- Add 2 entries inside assertioninfor.property as following:
----saml.DigestAlgorithm=http://www.w3.org/2000/09/xmldsig#sha1
----saml.SignatureAlgorithm=http://www.w3.org/2000/09/xmldsig#rsa-sha1

Note: Plan to consolidate lookup for multiple sha as following:
SHA.outbound=SHA1,SHA256,SHA512
SHA.inbound=SHA1,SHA256,SHA512

### DIFF
--- a/Product/Production/Gateway/PatientDiscovery_10/src/main/resources/patientdiscovery/_10/webservices.xml
+++ b/Product/Production/Gateway/PatientDiscovery_10/src/main/resources/patientdiscovery/_10/webservices.xml
@@ -6,9 +6,6 @@
     in writing, software distributed under the License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
     ANY KIND, either express or implied. See the License for the specific language governing permissions and limitations under
     the License. -->
-    
-
-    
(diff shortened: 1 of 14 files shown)


Product/Production/Gateway/PatientDiscovery_10/src/main/resources/patientdiscovery/_10/webservices.xml
```diff
@@ -13,27 +13,33 @@
...
 
+    
     <!-- Patient Discovery -->
 
     <cxf:bus>
         <cxf:features>
             <p:policies />
         </cxf:features>
     </cxf:bus>
+    <context:property-placeholder />
```
plobre, line 33: Just some left over cleanup, This was when we were trying to inject the jaxws properties into the endpoints. This is no longer needed and can be removed, along with the schemaLocations above.

PR #1846, opened 2019-01-10
Conn 2383 Pick certificate alias by Exchange Name instead of URL

plobre:

PR #1851, opened 2019-01-15
implement SNI based on exchange

mhpnguyen:


mhpnguyen:
retest this please

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/messaging/service/decorator/cxf/SSLSocketFactoryWrapper.java
```diff
@@ -0,0 +1,109 @@
...
+    private final SSLSocketFactory wrappedFactory;
+    private final SSLParameters sslParameters;
+
+    public SSLSocketFactoryWrapper(SSLSocketFactory factory, SSLParameters sslParameters) {
+        wrappedFactory = factory;
+        this.sslParameters = sslParameters;
+    }
+
+    @Override
+    public Socket createSocket(String host, int port) throws IOException, UnknownHostException {
```
connectscagitbot: MINOR Remove the declaration of thrown exception 'java.net.UnknownHostException' which is a subclass of 'java.io.IOException'. [rule](http://192.168.3.101:9000/coding_rules#rule_key=squid%3ARedundantThrowsDeclarationCheck)

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/messaging/service/decorator/cxf/SSLSocketFactoryWrapper.java
```diff
@@ -0,0 +1,109 @@
...
+        SSLSocket socket = (SSLSocket) wrappedFactory.createSocket(host, port);
+        setParameters(socket);
+        return socket;
+    }
+
+    @Override
+    public Socket createSocket(String host, int port, InetAddress localHost, int localPort) throws IOException,
+    UnknownHostException {
```
connectscagitbot: MINOR Remove the declaration of thrown exception 'java.net.UnknownHostException' which is a subclass of 'java.io.IOException'. [rule](http://192.168.3.101:9000/coding_rules#rule_key=squid%3ARedundantThrowsDeclarationCheck)

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/cryptostore/StoreUtil.java
```diff
@@ -43,7 +44,7 @@
 public class StoreUtil {
 
     private static final Logger LOG = LoggerFactory.getLogger(StoreUtil.class);
-    public static final String INTERNAL_EXCHANGE = "INTERNAL";
+    public static final String INTERNAL_EXCHANGE = InternalExchangeManager.getInstance().getDefaultExchange();
```
plobre: Rather than initialize this as a final static variable, Id rather call this method each time. If the internalExchangeInfo.xml file is updated, we lose the ability to reload the file and update this field.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/exchangemgr/InternalExchangeManager.java
```diff
@@ -123,7 +123,7 @@ public OrganizationType getOrganization(String hcid) throws ExchangeManagerExcep
 
     @Override
     public String getDefaultExchange() {
-        return null;
+        return null != exInfo ? exInfo.getDefaultExchange() : null;
```
plobre, line 134: Rather than just a null check, shouldn't we attempt to load the exchange info file first?

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/messaging/client/CONNECTCXFClientFactory.java
```diff
@@ -64,7 +69,10 @@
         if (StringUtils.isEmpty(url)) {
             throw new WebServiceException(ERROR);
         }
-
+        if (StringUtils.isBlank(target.getExchangeName())) {
```
plobre, line 72: Im handling this logic already in `CONNECTCXFClientSecured`. I think we should just update the logic already there. Additionally the other getter here (without the NhinTargetSystemType) above on line 51 also needs to make use of this change.

See https://github.com/CONNECT-Solution/CONNECT/pull/1851/files#diff-c7700e56eac1a9d3585fc702a81e867eR79

plobre (changes requested):

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/callback/cxf/CXFSAMLCallbackHandler.java
```diff
@@ -92,7 +94,10 @@ public void handle(final Callback[] callbacks) throws IOException, UnsupportedCa
                     final Message message = getCurrentMessage();
                     AssertionType custAssertion = getCustAssertion(message.get("assertion"));
                     NhinTargetSystemType systemType = (NhinTargetSystemType) message.get("targetSystemType");
-                    String exchangeName = systemType != null ? systemType.getExchangeName() : StoreUtil.INTERNAL_EXCHANGE;
+                    String exchangeName = systemType != null ? systemType.getExchangeName() : InternalExchangeManager
```
TabassumJafri, line 97: Why we are using InternalExchangeManager here? We have defaultExchange in exchangeInfo as well. Are they going to be the same thing. If yes we need to document this and possibly update regression suites

TabassumJafri (changes requested):

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/callback/cxf/CXFSAMLCallbackHandler.java
plobre: They shouldn't be the same thing. This was the issue I originally ran into with #1846 . Entity and Adapter layers don't have the NhinTargetSystemType within the scope of when we make the CONNECT client calls. In this case, `systemType` would be NULL here. Since these two layers are within the Internal exchange, I had added that constant `INTERNAL_EXCHANGE` to handle this case. Minh has changed it so we can now pull from the internal exchange info instead of having a placeholder constant for those 2 layers.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/messaging/client/CONNECTCXFClientFactory.java
mhpnguyen: The logic here is different.  We want to handle the default exchange if empty since our NHIN layer only call this method.

Product/Production/Common/CONNECTCoreLib/src/main/java/gov/hhs/fha/nhinc/callback/cxf/CXFSAMLCallbackHandler.java
mhpnguyen: We handle this in case for secure adapter scenario

plobre (approved):

TabassumJafri (approved):

ptambellini01 (approved):

trantang (approved):

PR #1903, opened 2019-05-03
Conn 2454 - Deferred Response Option Query

plobre:
Adds the IG's Deferred response option query to store the request ID, DC ID, and the desired endpoint to respond to.


Product/Production/Gateway/DocumentQuery_30/src/main/java/gov/hhs/fha/nhinc/docquery/_30/entity/DeferredDocQueryCheck.java
```diff
@@ -0,0 +1,104 @@
...
+import oasis.names.tc.ebxml_regrep.xsd.query._3.AdhocQueryResponse;
+import oasis.names.tc.ebxml_regrep.xsd.rs._3.RegistryError;
+import oasis.names.tc.ebxml_regrep.xsd.rs._3.RegistryErrorList;
+import org.apache.commons.lang.StringUtils;
+
+/**
+ * @author ptambellini
+ *
+ */
+public final class DeferredDocQueryCheck extends BaseService {
```
TabassumJafri: Why we are extending this class from BaseService?

Product/Production/Services/DocumentQueryCore/src/main/java/gov/hhs/fha/nhinc/docquery/deferred/impl/AdapterDeferredResponseOptionImpl.java
```diff
@@ -0,0 +1,54 @@
...
+        AdhocQueryRequest msg,
+        AssertionType assertion) {
+
+        String newId = saveRequest(msg, assertion);
+        AdapterDeferredResponseOptionQuerySecuredType response = new AdapterDeferredResponseOptionQuerySecuredType();
+        response.setNewId(newId);
+        return response;
+    }
+
+    private static String saveRequest(AdhocQueryRequest msg, AssertionType assertion) {
```
TabassumJafri: Optional: this method is just confusing. We are calling it SaveRequest but a new Id is generated and returned here.

TabassumJafri (changes requested):

trantang (approved):

ptambellini01 (approved):

TabassumJafri (approved):

